PR #6, opened 2026-01-16
Clean up of the DND plugin itself, and related ScriptUtils

xulman:
ScriptUtils:
- util function to read the user-provided title of its preset script
- simplified calling of the scripts (skipped explicit ModuleService)

DND plugin:
- removed the earlier (and not used) solution and its codebase
  (e.g. welcome opening dialog, detection of the ALT key,...)
- more concise reporting
- reads (and shows) the user-provided title of its preset script


stefanhahmann (approved):
Got it.
Good that unused stuff is cleaned up with this PR.

PR #37, opened 2026-03-06
Open specific resolution level in case of setting open matching IJ resolution level

stefanhahmann:
This pull request introduces support for loading OME-Zarr images at a preferred width, allowing users to avoid loading images at full resolution if they are too large for memory.

### Preferred resolution support for OME-Zarr image loading

* Added a new constructor parameter `preferredMaxWidth` to `DefaultPyramidal5DImageData`, and implemented logic in `selectResolutionLevel` to choose the best resolution level based on this value. If the highest resolution exceeds the preferred width, a downsampled version is chosen; otherwise, the highest available resolution is used.
* Modified `ZarrOpenActions` methods and plugin logic to pass the `preferredMaxWidth` parameter when opening images with preferred resolution, ensuring the new feature is integrated with the drag-and-drop handler and image opening workflows.

### Test enhancements

* Updated and added tests to verify that images are loaded at the correct resolution based on the `preferredMaxWidth`.

### DIFF
--- a/src/main/java/sc/fiji/ome/zarr/ui/ZarrDragAndDropOpenSettingsUI.java
+++ b/src/main/java/sc/fiji/ome/zarr/ui/ZarrDragAndDropOpenSettingsUI.java
@@ -5,6 +5,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import org.scijava.ItemVisibility;
 import org.scijava.command.Command;
 import org.scijava.command.DynamicCommand;
 import org.scijava.plugin.Parameter;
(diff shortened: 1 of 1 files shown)


src/main/java/sc/fiji/ome/zarr/util/ZarrOpenActions.java
```diff
@@ -134,11 +141,11 @@ private void showNonZarrError( final Exception e )
 	}
 
 	void openImage( final Consumer< PyramidalDataset< ? > > multiScaleImageOpener, final Consumer< Img< ? > > singleScaleImageOpener,
-			final String message )
+			final String message, final Integer preferredWidth )
 	{
 		try
 		{
-			openMultiScaleImage( multiScaleImageOpener );
+			openMultiScaleImage( multiScaleImageOpener, preferredWidth );
```
xulman: Seeing the "flood" of `preferredWidth` params in methods, I'm wondering if it wouldn't be better to have this as an attribute of this class. The class is anyway instantiated anew with every new dropped image, so it is "sensitivite to"/is_picking_up_fresh current values of the preferred width.

PR #57, opened 2026-04-13
Update N5 dependency versions in POM and README files

stefanhahmann:
* Affects n5, n5-hdf5, n5-imglib2, n5-universe, n5-zarr


xulman:
Many thanks for updating to the current latest deps.

I checked the versions list in README and pom.xml, and they do match. yay! :)

xulman (approved):

PR #59, opened 2026-04-13
Add GitHub Actions workflow for cross-OS testing on push and pull requests

stefanhahmann:
This PR adds a new GitHub Actions workflow for cross-platform testing. The main focus is on ensuring tests run correctly on Linux, macOS, and Windows, with appropriate setup for platform-specific dependencies.

**Continuous Integration Enhancements:**

* Added a new workflow file `.github/workflows/test-os.yml` to run tests on `ubuntu-latest`, `macos-latest`, and `windows-latest` runners, including platform-specific installation steps for the Blosc library.

**Test Configuration Improvements:**

* Updated the Maven Surefire plugin configuration in `pom.xml` to explicitly set `<forkCount>0</forkCount>` (disabling forking), add `-Dimagej.legacy.disable=true` to the JVM arguments, and retain existing memory and assertion settings.

### DIFF
--- a/.github/workflows/test-os.yml
+++ b/.github/workflows/test-os.yml
@@ -1,0 +1,61 @@
+name: test-os
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*-[0-9]+.*"
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  mvn-verify:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Java
+        uses: actions/setup-java@v4
+        with:
+          java-version: '11'
+          distribution: 'zulu'
+          cache: 'maven'
+
+      # Linux-only setup
+      - name: Linux dependencies (Blosc via apt)
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y xvfb libblosc-dev
+          Xvfb :99 -screen 0 1280x1024x24 &
+          echo "DISPLAY=:99" >> $GITHUB_ENV
+
+      # macOS setup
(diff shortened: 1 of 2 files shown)


xulman (approved):
The new .yml looks reasonable: The same jdk11 base, complemented with platform-specific recipe to fetch and make-ready blosc (btw, kudos for the vcpkg way for the win platform... that's indeed how development stuff shall be done for the c++ land today; ...poor GH servers....).

xulman:
Only a couple of curious questions, or just dbl-checking review, to understand better:

The `workflow` files, they are all triggered on the same occasions :heavy_check_mark:, and their purpose is:
- `build.yaml` to build on Linux and publish to maven central?
   - only this one refers to both `.github/*sh` files
   - using which it builds *in the scijava-way*, which probably
     - assures compatibility with Fiji
     - deploys to maven central eventually
- `test-os.yaml` to build on three platforms and only run tests (via `mvn -B verify`)?
- `sonarcloud.yaml` to check with SonarCloud... but only here with jdk**17** (everywhere else we go with **11**)

stefanhahmann:
> Only a couple of curious questions, or just dbl-checking review, to understand better:
> 
> The `workflow` files, they are all triggered on the same occasions ✔️, and their purpose is:
> 
>     * `build.yml` to build on Linux and publish to maven central?

Maven central, I am not sure. But it publishes to _maven.scijava.org_.
       
>       * only this one refers to both `.github/*sh` files
>       * using which it builds _in the scijava-way_, which probably
>         
>         * assures compatibility with Fiji
>         * deploys to maven central eventually

Yes, this the _scijava-way_.

> 
>     * `test-os.yaml` to build on three platforms and only run tests (via `mvn -B verify`)?

Yes, exactly. This ensures multi-platform compatibility
> 
>     * `sonarcloud.yaml` to check with SonarCloud... but only here with jdk**17** (everywhere else we go with **11**)

The JDK 17 is a minimum requirement of the SonarCloud job. For the remaining jobs, there is actually no super good reason why to choose JDK 11. We could change that in the future.

PR #61, opened 2026-04-14
Change mvn build profile

stefanhahmann:
This PR updates test dependencies. Test dependencies are changed from `imagej` to `fiji` (excluding `imagej-legacy`)
The maven profile `ui-tests` is removed, since it is not needed anymore.
Extra to this it is ensured all windows are disposed at the end of `testRunScriptWithScriptSpecified` in `ZarrOpenActionsTest.java`.

This PR became possible, since the Java heap space error observed on Macos earlier when running lots of unit tests in a row could be narrowed down to the simulatanoues existence of `scijava-common` and `imagej-legacy`: https://github.com/fiji/fiji/issues/427#issuecomment-4178009760

Ideally should only be merged after https://github.com/BioImageTools/ome-zarr-fiji-java/pull/59 has been merged to allow testing this on macos in the ci pipeline.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,12 +194,17 @@
 		<!-- ====================================== -->
 		<!-- test dependencies -->
 		<!-- ====================================== -->
-
-		<!-- for testing/development purposes: include imagej-legacy at test runtime -->
+		<!-- for testing/development purposes: include imagej at test runtime -->
 		<dependency>
-			<groupId>net.imagej</groupId>
-			<artifactId>imagej</artifactId>
+			<groupId>sc.fiji</groupId>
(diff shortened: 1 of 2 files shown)


xulman (approved):
I like the proposed solution better compared to the current/previous solution.

While the previous solution enabled the full `fiji.sc` as well, all dependencies were not listed in the same `<dependency>` section of the `pom.xml`, and the `<dependency>` section mentioned `net.imagej` instead (a false impression at the first glimpse).

The proposed solution lists `fiji.sc` directly (that's what one normally wants) and excludes the IJ1 subpart only for the tests (because of the *context issue*).

PR #63, opened 2026-04-16
Fix bug while running scripts

stefanhahmann:
This PR fixes a bug that occured when running scripts on imported OME-Zarrs

**Error handling and messaging improvements:**

* Added explicit error messages when required services (`ScriptService` or `PrefService`) are unavailable, clarifying potential issues with Fiji initialization (`ScriptUtils.java`).

**Script execution robustness:**

* Ensured that the `ScriptModule` context is explicitly set before script execution, which is important for proper script functioning within the Fiji environment (`ScriptUtils.java`).

**Testing enhancements:**

* Updated tests to check for script failures, improving test reliability and coverage of error scenarios (`ZarrOpenActionsTest.java`).
* Modified test setup for single scale image opening methods (actually unrelated to the bug)

**Dependency management:**

* Removed unnecessary exclusions from the `fiji` test dependency in `pom.xml`, since the Java heap space error is not occurring any longer.

### DIFF
--- a/src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
+++ b/src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
@@ -107,6 +107,7 @@ public class ScriptUtils
 			try
 			{
 				ScriptModule module = scriptService.getScript( new File( scriptPath ) ).createModule();
+				module.setContext( ctx );
 				module.setInput( "path", inputPath );
 				logger.info( "Executing script: {}", scriptPath );
 				logger.info( "on String parameter: {}", inputPath );
--- a/src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
+++ b/src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
@@ -129,7 +129,7 @@ public class ScriptUtils
 		}
 		else
 		{
-			logger.debug( "Script path is not valid: {}. Opening script editor with template.", scriptPath );
+			logger.info( "Script path is not valid: {}. Opening script editor with a script template instead.", scriptPath );
 			//this opens an _always new_ window with the template script,
 			//...at least the user is more likely to notice that this "help" came up
 			final TextEditor editor = new TextEditor( ctx );
--- a/src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
+++ b/src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
@@ -130,6 +130,8 @@ public class ScriptUtils
 		else
 		{
 			logger.info( "Script path is not valid: {}. Opening script editor with a script template instead.", scriptPath );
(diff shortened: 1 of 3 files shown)


src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
```diff
@@ -93,6 +93,8 @@ public static void executePresetScript( final Context ctx, final String inputPat
 		PrefService prefService = ctx.getService( PrefService.class );
 		if ( scriptService == null || prefService == null )
 		{
+			errorHandler.accept(
+					"Service for running scripts and/or service for reading preferences are not available. Is Fiji properly initiated?" );
```
xulman, line 98: Maybe this `errorHandling` could also happen after the `logger` message is fired, just like below?

src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
xulman: practical comment from testing:

When I preset an incorrect path to a script (or when the script moved after the path to it had been preset), the user isn't notified (because of a `debug` here) and is surprised instead to see the MacroEditor with the example script --- which is good, I would like to keep it; just perhaps we could **also somehow inform** the user that the script is not valid (*).

(*) An idea: WHen the script execution fails of not finding the script, and dialog informing about it is poped out. That dialog could have two buttons: Start the ScriptPreset plugin, Start the text editor with the example. Perhaps this could be left to some next PR.... but since we're already improving the scripting here already ... :-)

xulman (dismissed):
Tested on Linux, works well.

I have left an (non-critical) idea for maybe one or two more commits to this PR :-)

src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
stefanhahmann: Done.

src/main/java/sc/fiji/ome/zarr/util/ScriptUtils.java
stefanhahmann: I decided to increase the log level to info. By this, at least something is printed to the console.

I decided against the two options in the info dialog, since the first option would require too much explanation about what actually will happen, when this option is chosen by the user. Also, it may change in the future and then we would need to adapt this as well.

The dialog however is a good point. I added it. Also gave the posibility to improve the associated unit test.

@xulman could you please test the user experience once more?

xulman: Okay, the implemented solution is also a very nice and good one (just fixed a typo in the report).

The `.info()` isn't still showing on my console, but the dialog comes out. So, it is probably okay that the (duplicit) `info` is silent.



Tested on WinX and Linux, works well everywhere.

xulman: Unless we want to fix the `info()` not showing up, I would merge already.

PR #69, opened 2026-04-27
Add zarr-java as an alternative OME-Zarr reader backend

stefanhahmann:
## Summary                                                                                                                                                                                                                       
                                                                                                                                                                                                                                   
  Adds [`dev.zarr:zarr-java`](https://github.com/zarr-developers/zarr-java) as a second OME-Zarr reader backend alongside the existing N5-universe backend. Users can pick which library reads their datasets from **Plugins →     
  OME-Zarr → Drag & Drop Behavior Settings**; N5 stays the default. The N5 loading code is extracted into a small `PyramidBackend` SPI, the new zarr-java implementation lives next to it, and `Pyramidal5DImageData` gains static
  factories so consumers don't have to import the impl/backends directly.                                                                                                                                                          
                                                                                                                                                                                                                                 
  ## What's in here

  The 13 commits are arranged so each one is a self-contained, reviewable unit (and each builds in isolation):

  1. **Remove license headers from v5 `zarr.json` test resources** — strict JSON parsers reject the embedded comment-style header.                                                                                                 
  2. **Add zarr-java dependency to `pom.xml`** — `dev.zarr:zarr-java:0.1.2`
  3. **Move pyramid exception classes into a dedicated `exceptions` sub-package** — pure rename + import updates.                                                                                                                  
  4. **Polish Omero metadata** — rename `name` → `version` (matches OME-Zarr spec) and hoist a shared `buildChannelLabels` helper onto `Omero` so both backends can use it.                                                        
  5. **Extract N5 loading into a `PyramidBackend` abstraction** — pure refactor introducing `PyramidBackend` / `PyramidContents` / `N5PyramidBackend`; `DefaultPyramidal5DImageData` becomes the backend-agnostic                  
  `Pyramidal5DImageDataImpl`. No behavior change.                                                                                                                                                                                  
  6. **Add zarr-java backend implementation** — `ZarrJavaPyramidBackend` + `ZarrJavaCellLoader`.                                                                                                                                   
  7. **Add `ZarrReaderBackend` enum and persist it in drag-and-drop settings** — `N5` (default) and `ZARR_JAVA`, read/written via `PrefService`.                                                                                   
  8. **Switch backend in `ZarrOpenActions`** — honors the configured setting on drag-and-drop opens.                                                                                                                               
  9. **Expose backend selection in the drag-and-drop settings UI** — adds the choice field to `DragAndDropBehaviorSettings`; updates supported-version text to v0.4 / v0.5.                                                        
  10. **Add static factory methods to `Pyramidal5DImageData`**:                                                                                                                                                                    
      - `Pyramidal5DImageData.open(context, path, preferredWidth)` — uses the configured backend
      - `Pyramidal5DImageData.openWithN5(context, path, preferredWidth)` — forces N5                                                                                                                                               
      - `Pyramidal5DImageData.openWithZarrJava(context, path, preferredWidth)` — forces zarr-java
  11. **Refactor pyramid tests as a JUnit 5 interface and split per backend** — `Pyramidal5DImageDataTestBase` becomes a JUnit 5 test interface with `default` test methods; `N5BackedPyramidal5DImageDataTest` and                
  `ZarrJavaBackedPyramidal5DImageDataTest` are thin classes that supply the backend-specific factory. Both run the full shared suite, and IntelliJ shows runnable green arrows on the concrete classes.                            
  12. **Add `OmeroMetadataAccessTest`** — lomero metadata access test using zarr-java for v0.4 and v0.5.                                                                                                                             
  13. **Add `BackendBenchmark`** — `main()`-driven benchmark to compare load and random-access timings between the two backends; not part of the regular test suite.                                                               
                                                                                                                                                                                                                                   
  ## Test plan
                                                                                                                                                                                                                                   
  - [x] `mvn clean test` passes — 473 tests, 0 failures, 0 errors (with c-blosc on `DYLD_LIBRARY_PATH`).                                                                                                                           
  - [x] Both backends run the full 194-test parameterized suite over the same v0.4 / v0.5 fixtures.
  - [x] Drag-and-drop an OME-Zarr v0.4 dataset into Fiji with each backend selected.                                                                                                                                               
  - [x] Drag-and-drop an OME-Zarr v0.5 dataset into Fiji with each backend selected.                                                                                                                                               
  - [x] Verify the backend choice persists across Fiji restarts.                                                                                                                                                                   
  - [x] Verify the N5 default behavior is unchanged for users who never touch the new setting.


src/main/java/sc/fiji/ome/zarr/pyramid/metadata/Omero.java
```diff
@@ -89,4 +95,18 @@ public String toString()
 	{
 		return new Gson().toJson( this );
 	}
-}
+
+	public static String[] buildChannelLabels( final String fallbackName, final Omero omero, final int numChannels )
+	{
+		final boolean omeroValid = omero != null && omero.channels != null && omero.channels.size() == numChannels;
```
xulman, line 101: I remember this code from earlier... and the commit says that it has been hoisted.. but here it appears, must be thus missing elsewhere, where?

src/main/java/sc/fiji/ome/zarr/pyramid/metadata/Omero.java
```diff
@@ -28,20 +28,26 @@
...
 @SuppressWarnings( "all" )
 public class Omero
 {
+	private static final Logger logger = LoggerFactory.getLogger( MethodHandles.lookup().lookupClass() );
+
 	// Top-level Omero class
 	public int id;
 
-	public String name;
+	public String version;
```
xulman: Haven't we changed this before to `name`, I recall some fix to match the OMERO specs in NGFF?

xulman: Indeed, check the commit 20c700f5848ce90e1984c995a7f8b0e2443d35d8

xulman: I don't get that it works with `version` :)

Even [the latest 0.6dev spec suggests name](https://ngff.openmicroscopy.org/specifications/dev/index.html#omero-metadata-transitional)

Did Claude say something about it in its log?

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
```diff
@@ -0,0 +1,309 @@
...
+ * The image data is produced by a {@link PyramidBackend} (e.g.
+ * {@link N5PyramidBackend}); this class copies the backend's output into its
+ * own state, builds the ImageJ {@link Dataset} and the BigDataViewer
+ * {@link SourceAndConverter} list, and exposes everything through
+ * {@link Pyramidal5DImageData}.
+ *
+ * @param <T> pixel type
+ * @param <V> volatile pixel type
+ */
+public class Pyramidal5DImageDataImpl<
```
xulman, line 78: Hmm... I'm a little hesitant about the `...Impl` vs. `Default....` naming schemes.

The latter is, for sure, consistent with scijava... but perhaps less modern?
IDK, for historical reasons, I would maybe suggest going with the `DefaultPyramidal5DImageData`.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
```diff
@@ -0,0 +1,282 @@
...
+import net.imglib2.cache.img.CachedCellImg;
+import net.imglib2.realtransform.AffineTransform3D;
+import net.imglib2.type.NativeType;
+import net.imglib2.type.numeric.RealType;
+
+import mpicbg.spim.data.sequence.VoxelDimensions;
+import sc.fiji.ome.zarr.pyramid.metadata.Omero;
+
+/**
+ * Immutable snapshot of everything a {@link PyramidBackend} produces when
```
xulman, line 43: Can we use `record` from later versions of Java?

Would it work with Fiji stable (I guess not) -> Do we want to support *Fiji-stable*?
(Despite it being called *stable*, I read it as legacy, historical relic for moments when *Fiji-latest* isn't working...)

And if `record` would work, we might even consider to not copy but use directly this content in `Pyramidal5DImageDataImpl`.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
```diff
@@ -0,0 +1,282 @@
...
+
+	public final VoxelDimensions voxelDimensions;
+
+	public final AffineTransform3D[] transforms;
+
+	public final CachedCellImg< T, ? >[] cachedCellImgs;
+
+	public final RandomAccessibleInterval< V >[] volatileImgs;
+
+	public final ImgPlus< T > imgPlus;
```
xulman, line 83: Do we still need this here?

xulman: We have it here only because of the `DefaultPyramidal5DImage`, but

a) can't `DefaultPyramidal5DImage` built it there on its own?
b) does `DefaultPyramidal5DImage` need it in the end? :)

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
```diff
@@ -0,0 +1,282 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+package sc.fiji.ome.zarr.pyramid.backend;
+
+import net.imagej.ImgPlus;
```
xulman, line 31: Something possibly worth considering:

----

Because of this nice refactoring, I'm now very much tempted to offer this `backend` package as an artifact on its own, an artifact that bridges an OME-Zarr dataset to a Java class that represents it. This class (`PyramidContents`) could be made to depend only on `imglib2`  (which is okay for Icy and QuPath). Then, our current `Pyramidal5DImageData` would be a wrapper to specific for Fiji, exposing pixels as BDV's `SAC`, ImageJ2's `Dataset` and ImageJ's `ImagePlus`. And others (Icy, QuPath) could wrap with their own "pixel data" classes. And they could depend only on this class w/o pulling the fiji stuff/deps (like, e.g., `Dataset`).

Let's leave this for another PR. ...we could also lift up the `selectedResLevel` from backends...

src/main/java/sc/fiji/ome/zarr/settings/ZarrDragAndDropOpenSettings.java
```diff
@@ -119,9 +148,21 @@ public static ZarrDragAndDropOpenSettings loadSettingsFromPreferences( final Pre
 						ZarrDragAndDropOpenSettings.class, ZARR_PREFERRED_WIDTH_SETTING_NAME,
 						DEFAULT_PREFERRED_WIDTH
 				);
+		ZarrReaderBackend backend;
+		try
+		{
+			backend = prefs == null ? DEFAULT_READER_BACKEND : ZarrReaderBackend
+					.getByName(
+							prefs.get( ZarrDragAndDropOpenSettings.class, ZARR_READER_BACKEND_SETTING_NAME, DEFAULT_READER_BACKEND.name() ) );
```
xulman: the line breaking by Claude is weird (looks terrible in Git or VIM, for example),
IntelliJ mixes tabs and spaces that I like better :-)    (and looks normal in Git and VIM)

xulman (dismissed):
Many thanks. Looks good, and triggered a lot of inspiration again...

I didn't check if the copied codes are copied verbatim...

src/main/java/sc/fiji/ome/zarr/pyramid/metadata/Omero.java
stefanhahmann: It was the human (i.e. me) who made the mistake, not the AI.

I saw `version` here: https://github.com/zarr-developers/zarr-java/blob/main/src/main/java/dev/zarr/zarrjava/experimental/ome/metadata/OmeroMetadata.java#L22
I did not see that there is `name` as well. Thus I thought this has to be changed and instructed Claude to do so.
I did not double check with the Spec though and that was the fault. Thanks for catching this!

I am wondering, why there is `version` at all in here: https://github.com/zarr-developers/zarr-java/blob/main/src/main/java/dev/zarr/zarrjava/experimental/ome/metadata/OmeroMetadata.java

Maybe this field existed in earlier version of the spec...

I changed it back to name.

src/main/java/sc/fiji/ome/zarr/pyramid/metadata/Omero.java
stefanhahmann: The logic used to live here

- https://github.com/BioImageTools/ome-zarr-fiji-java/blob/main/src/main/java/sc/fiji/ome/zarr/pyramid/DefaultPyramidal5DImageData.java#L263-L267

and here:

- https://github.com/BioImageTools/ome-zarr-fiji-java/blob/main/src/main/java/sc/fiji/ome/zarr/pyramid/DefaultPyramidal5DImageData.java#L282

I moved it, because both backends needed to do the same.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
stefanhahmann: In order to support multiple backends, I was wondering, which strategy would be good.

One way was having: Multiple implementations of `Pyramdial5DImageData`, e.g. `N5PyramidalImageData` and `ZarrJavaPyramidalImageData` with an `AbstractPyramidalImageData` in between. 

The I asked Claude, if this is a case of _composition over inheritance_ or not. It had a slight tendency towards having only one implementation of Pyramidal5DImageData that supports different backends. With this strategy more backends could be added a bit easier in the future (however, not sure, if there ever will be more than these two).

Since for `Pyramidal5DImageData`, there really is now only one implementation of the interface, I decided for `...Impl` rather than `Default...`. One could even argue that the separation of the interface `Pyramidal5DImageData` and its implementation `Pyramidal5DImageDataImpl` is not required any longer. I kept the interface for historical reasons.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
stefanhahmann: Good idea. I haven't used `record` so far, but it seems to have been invented for use cases like `PyramidContents`.

At the moment, FIJI stable is still mostly supported by our plugin. In the known issues of our readme, we indicate, that it can be used with FIJI stable, except of blosc compression and Mac.

grafik

Since, I still sometimes use `stable` (and could imagine more conservative users also still using it), I am still a bit reluctant to go for Java > 8 language features..

src/main/java/sc/fiji/ome/zarr/settings/ZarrDragAndDropOpenSettings.java
stefanhahmann: Indeed.
This is an artefact of letting Claude doing the commits as well. It then does not seem to use the formatting settings of IntelliJ anymore.
I reformatted the code with the IntelliJ settings.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
stefanhahmann: Good idea, indeed.
I created https://github.com/BioImageTools/ome-zarr-fiji-java/issues/72

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
stefanhahmann: Probably not.
This is a leftover that got carried over to PyramidContents. I would like to leave this to a separate PR, if you don't mind: https://github.com/BioImageTools/ome-zarr-fiji-java/issues/73

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
xulman: I see, no problem with it.

(It would have perhaps made our life easier (and a code a little nicer) after we detach from the ancient JVM. It needs time...)

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
xulman: OKay, I take both, I'm happy with it in the end. In fact, the ony thing that triggered my attention was this naming.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
stefanhahmann: I agree. I would also like to use Java 9+ features sometimes, but support for FIJI stable still feels at least a bit relevant for me.

PR #74, opened 2026-05-05
Add Uri support

### LINKED ISSUE #28: DefaultPyramidal5DImageData could be constructed on URL, not only on paths.
This is a bookmark issue to remind ourselves of the observation that `DefaultPyramidal5DImageData` objects could, in principle, be constructed from general URLs, not necessarily only from [filesystem paths as it is now](https://github.com/BioImageTools/ome-zarr-fiji-java/blob/d392e186c138bc6d1d17ee80c8f420041937135e/src/main/java/sc/fiji/ome/zarr/pyramid/DefaultPyramidal5DImageData.java#L169).

stefanhahmann:
Resolves https://github.com/BioImageTools/ome-zarr-fiji-java/issues/28
                                                                                                                                                                                                                                                                                                                                                                     
##  New features for users                                                                                                                                                                                                           
                                                                                                                                                                                                                                   
* Open remote OME-Zarr datasets over HTTP(S)                                                                                                                                                                                       
  * All opening paths now accept https:// and http:// URLs in addition to local file-system paths.
  * Both the N5 and zarr-java backends handle remote URLs.
  * Paste an OME-Zarr URL from the clipboard with three entry points:                                                                                                                                                                    
    * IJ1 toolbar button: a one-shot paste icon sits in the legacy ImageJ toolbar (uses a non-breaking hyphen in the name to avoid IJ1 parsing issues)
    * Plugins menu command: "Paste OME-Zarr URL" is registered as a SciJava command under the Plugins menu                                                                                                                           
    * Keyboard shortcut: Ctrl/Cmd/Shift+V fires from any ImageJ window
      * Accepts zarr locations
      * Falls back to default for non zarr locations / images
      * Suppressed when a text component such as the script editor has focus
* Pasting a plain local file-system path from the clipboard also works.
* Clearer error for bioformats2raw multi-image collections
  * When a dropped or pasted path points to the root of a bioformats2raw-produced collection, the plugin now detects the bioformats2raw.layout marker and reports a more targeted message.
                                                                                                
##  API changes (developer-facing)
- `Pyramidal5DImageData(String path, …)` → `Pyramidal5DImageData(URI uri, …)  `                                                                                                                                                      
- `ZarrOpenActions(String path, …)` → `ZarrOpenActions(URI uri, …)` and moved to new package `sc.fiji.ome.zarr.open`
- `ZarrDragAndDropOpenSettings` renamed to `ZarrOpeningSettings`
 
##  Internal improvements

- new class `ZarrLocations` — classifies a URI as local path, remote URL, or unrecognised; shared by both backends and the HTTP detection logic
- new class `ClipboardUtils` — parses a valid OME-Zarr URI from the system clipboard; used by all three paste entry points
- new class `ZarrOverHttpUtils` — HEAD-request probe to verify a remote Zarr root is reachable before opening
- new method `ZarrOpenActions.openWithSettings` — static method that reads stored preferences and dispatches to the correct open action; used by `DnDHandlerPlugin` and the 3 Paste entry points
- `MultiImageDatasetException` — dedicated exception type for the bioformats2raw collection case
- `DnDActionChooser` — layout changed to 3×2 grid

##  Tests
       
- `PasteOmeZarrUrlToolInstallerTest` — lifecycle, keyboard dispatcher filtering, and toolbar action tool
- `ClipboardActionsTest` — paste action with valid/invalid clipboard content                                                                                                                                                       
- `ClipboardUtilsTest` — URI parsing from clipboard
- `ZarrLocationsTest` — path/URL classification                                                                                                                                                                                    
- `ZarrOverHttpUtilsTest` — HTTP detection and reachability probing                                                                                                                                                                
- `DnDActionChooserTest` — constructor and show/dispose lifecycle
- `ZarrOpenActionsTest` — extended: bioformats2raw collection handling parameterized over both backends (N5 and zarr-java)                                                                   
- bioformats2raw v5 test dataset added to test resources

## Manual Testing

###  HTTP(S) support
* [x] Open a remote https:// OME-Zarr URL via clipboard paste — image loads from network

### Toolbar button

* [x] Paste-arrow button appears in the IJ1 toolbar after Fiji starts                                                                                                                                                                
* [x] Click with valid URL in clipboard — dataset opens
* [x] Click with empty clipboard or non-URL text — FIJI clipboard opens
                                                                                                                                                                                                                                   
###  Plugins menu    
* [x]  Plugins → OME-Zarr → Open OME-Zarr from Clipboard is present and works                                                                                                                                                         
   
### Keyboard shortcut                                                                                                                                                                                                                
* [x] Ctrl/Shift/Cmd+V with IJ1 window focused — dataset opens
* [x] Ctrl/Shift/Cmd+V with script/macro editor focused — normal text paste, no OME-Zarr opener                                                                                                                                          
                                                                                                                                                                                                                               
### bioformats2raw collection error
* [x] Opening a bioformats2raw collection root shows a "multiple images" error                                                                                                                                                       
                                                                                                                                                                                                                                   
### Drag and drop regression
* [x]  Dragging a local OME-Zarr folder into Fiji still opens correctly


src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Detects whether a remote http(s) URL points at the root of an OME-Zarr
+ * dataset by issuing HEAD requests to the well-known Zarr metadata files.
+ */
+public class ZarrOverHttpUtils
+{
+	private static final Logger logger = LoggerFactory.getLogger( MethodHandles.lookup().lookupClass() );
+
+	private static final int CONNECT_TIMEOUT_MS = 5_000;
```
xulman: MS or MILLIS ?      :-)

My personal bias. But Google's AI thinks MS is more often used...

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
```diff
@@ -0,0 +1,126 @@
...
+import java.net.URI;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Detects whether a remote http(s) URL points at the root of an OME-Zarr
+ * dataset by issuing HEAD requests to the well-known Zarr metadata files.
+ */
+public class ZarrOverHttpUtils
```
xulman: Is this class only for HTTP? Or does it also handle S3, or any future protocols.... schemas (L80 below), in fact?
Why to restrict to http? (on the other hand, public s3 are typically channeled through http....)

And yes, method at L76 makes this http-only.

src/main/java/sc/fiji/ome/zarr/util/ZarrUtils.java
```diff
@@ -0,0 +1,126 @@
...
+				return false;
+			final String contentType = conn.getHeaderField( "Content-Type" );
+			return contentType == null || !contentType.toLowerCase().startsWith( "text/html" );
+		}
+		catch ( IOException e )
+		{
+			logger.debug( "HEAD request failed for {}: {}", uri, e.getMessage() );
+			return false;
+		}
+		finally
```
xulman, line 155: long time since I saw this :)
*try-with-resources* wouldn't work here?

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
```diff
@@ -0,0 +1,126 @@
...
+		if ( baseUri == null )
+			return false;
+		final String scheme = baseUri.getScheme();
+		if ( !"http".equalsIgnoreCase( scheme ) && !"https".equalsIgnoreCase( scheme ) )
+			return false;
+
+		final URI base = ensureTrailingSlash( baseUri );
+		for ( final String name : METADATA_FILES )
+		{
+			if ( existsByHead( base.resolve( name ) ) )
```
xulman: I've checked that both

```java
URI u = URI.create( "http://base/" );
System.out.println( u.resolve( "/subpath" ) );

URI u = URI.create( "http://base" );
System.out.println( u.resolve( "/subpath" ) );
```

create the same `http://base/subpath`

then either `METADATA_FILES` could be all with leading-'/', or here `resolve( '/'+name )`, and we could drop one function (`ensureTrailingSlash`) and make the code less verbose

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
```diff
@@ -0,0 +1,126 @@
...
+		return false;
+	}
+
+	private static URI ensureTrailingSlash( final URI uri )
+	{
+		final String s = uri.toString();
+		return s.endsWith( "/" ) ? uri : URI.create( s + "/" );
+	}
+
+	private static boolean existsByHead( final URI uri )
```
xulman: I know it's only a stupid util class, but still...

For this particular method, I had no idea what it was going to do. Could we rename to e.g. `canBeDownloaded()` or `canBeReached()`... or `isValidURI()` or `isFunctionalURI()` or alike....

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
xulman: Isn't a `file://foo` also a valid `URI`??

Thinking of merging somehow with `ZarrOnFileSystem` (which would probably need to be named better) to avoid code duplication. Btw, notice already the `ZarrOnFileSystem` is checking for presence of **four** files whereas here we check only **three**.... meaning a local valid OME-Zarr after moving to web could become invalid to our code (yes, that's very unlikely).

xulman (commented):
I've managed only a fraction so far. I'd like to continue ASAP.

With AI code, a lot of caution should be taken. AI prefers to generate (vast amounts of) code rather than looking for sleek solutions.

src/main/java/sc/fiji/ome/zarr/util/ZarrLocations.java
```diff
@@ -0,0 +1,73 @@
...
+		// prevent instantiation
+	}
+
+	/**
+	 * @param uri location to probe
+	 * @return {@code true} if the URI points at the root of a Zarr dataset.
+	 *   Returns {@code false} for unknown schemes, null, or any error during
+	 *   probing.
+	 */
+	public static boolean isZarr( final URI uri )
```
xulman: This one basically only unpacks `URI` to `Path`, checks that it is possible to do so, and then delegates to the class `ZarrOnFileSystemUtils`. Since it is also the only method in this util class, I would propose moving it to `ZarrOnFileSystemUtils` and dropping this class.

Besides, we have all util classes named as `*Util`, except for this one... yet another reason to lose it :-)

src/main/java/sc/fiji/ome/zarr/plugins/settings/OpeningBehaviorSettings.java
```diff
@@ -42,14 +42,14 @@
...
-import sc.fiji.ome.zarr.settings.ZarrDragAndDropOpenSettings;
+import sc.fiji.ome.zarr.settings.ZarrOpeningSettings;
 import sc.fiji.ome.zarr.settings.ZarrOpenBehavior;
 import sc.fiji.ome.zarr.settings.ZarrReaderBackend;
 
 /**
  * A FIJI/ImageJ command to select what to do when an OME-Zarr image is Drag &amp; Dropped into Fiji.
  */
-@Plugin( type = Command.class, menuPath = "Plugins > OME-Zarr > Drag & Drop Behavior Settings", initializer = "init" )
+@Plugin( type = Command.class, menuPath = "Plugins > OME-Zarr > Settings > Opening Behavior Settings", initializer = "init" )
```
xulman, line 52: I propose moving this class into a new package `sc.fiji.ome.zarr.plugins.settings` to mimic/show the menu layout in Fiji more closely. Easier to maintain and see the picture.

(On the other hand, I realize that the OME-Zarr Fiji submenu is somewhat temporarily collecting everything and may be re-arranged in the end... maybe??)

The same actually for `UserScriptSettings`.

BTW, also just like with the `UserScriptSettings` which dropped the `DragAndDrop` prefix, I would also rename this class to just `BehaviorSettings` or perhaps to `OpenBehaviourSettings`.

src/main/java/sc/fiji/ome/zarr/plugins/settings/UserScriptSettings.java
```diff
@@ -36,8 +36,8 @@
 
 import java.io.File;
 
-@Plugin( type = Command.class, menuPath = "Plugins > OME-Zarr > Drag & Drop User Script Settings", name = "OMEZarrDnDUserScript", headless = true )
-public class DragAndDropUserScriptSettings implements Command
+@Plugin( type = Command.class, menuPath = "Plugins > OME-Zarr > Settings > User Script Settings", name = "OMEZarrUserScript", headless = true )
+public class UserScriptSettings implements Command
```
xulman, line 40: I propose moving this class into a new package `sc.fiji.ome.zarr.plugins.settings` to mimic/show the menu layout in Fiji more closely. Easier to maintain and see the picture.

src/main/java/sc/fiji/ome/zarr/plugins/DnDHandlerPlugin.java
```diff
@@ -90,23 +88,9 @@ public Object open( final Location source ) throws IOException
...
-			break;
-		case BDV_MULTI_RESOLUTION:
-			actions.openBDVWithImage();
-			break;
-		case SHOW_SELECTION_DIALOG:
-		default:
-			createDnDActionChooser( context(), actions ).showDialog();
-			break;
-		}
+		ZarrOpenActions.openWithSettings( inputUri, context() );
```
xulman, line 92: hmm... that's actually nice

But then, perhaps, let's move the package `sc.fiji.ome.zarr.settings` under `open` so that `open` would just handle all kinds of opening, including those influenced by Fiji settings (hence this proposal for moving), etc.

I've studied what classes are using the content from the current `sc.fiji.ome.zarr.settings`, and it is mostly stuff from the `open` package. One exception is the default method `Pyramidal5DImageData.open()` from that iface.

src/main/java/sc/fiji/ome/zarr/settings/ZarrOpeningSettings.java
xulman: What if we were to move this package under `sc.fiji.ome.zarr.open` because, except for this one exception (`Pyramidal5DImageData.open()`), most of this package's content is consumed/used in the `open` package.

So it'd be nicely the *settings* of the *open*ing, which what it really is.

src/main/java/sc/fiji/ome/zarr/open/ClipboardActions.java
```diff
@@ -0,0 +1,42 @@
+package sc.fiji.ome.zarr.open;
```
xulman, line 1: thinking of the future write operations.. .shouldn't this package be better called `read` instead of `open`

as `open` ~ `close`, `read` ~ `write`
(but we could now end up with `open` ~ `write`)

src/main/java/sc/fiji/ome/zarr/pyramid/backend/n5/N5PyramidBackend.java
```diff
@@ -195,6 +195,27 @@ public PyramidContents< T, V > load()
...
+	{
+		final String scheme = inputUri.getScheme();
+		if ( scheme == null || "file".equalsIgnoreCase( scheme ) )
+		{
+			final Path inputPath = Paths.get( inputUri );
+			final Path rootPath = resolveRootPath( inputPath );
+			final List< String > elements = ZarrOnFileSystemUtils.relativePathElements( rootPath, inputPath );
+			return new ReaderTarget( rootPath.toUri().toString(), String.join( File.separator, elements ) );
+		}
+		if ( "http".equalsIgnoreCase( scheme ) || "https".equalsIgnoreCase( scheme ) )
```
xulman: This is again http-only, no?  Could we add S3 right away?

Or, do we need to restrict now at all?
Can't we outsource the failing (of not recognized transport protocol) later to N5 itself?

xulman (commented):
batch 2, some package moves suggestions

src/main/java/sc/fiji/ome/zarr/pyramid/backend/n5/N5PyramidBackend.java
stefanhahmann: I would like to add the S3 support in a different PR.

However, distinguishing the protocols was indeed not need. I remove it here: https://github.com/BioImageTools/ome-zarr-fiji-java/pull/74/changes/918b7fa14151a814e698c246b14a0ce164d9de4f

src/main/java/sc/fiji/ome/zarr/plugins/settings/UserScriptSettings.java
stefanhahmann: I agree. I did it as you suggested it.

src/main/java/sc/fiji/ome/zarr/plugins/settings/OpeningBehaviorSettings.java
stefanhahmann: I renamed to `OpeningBehaviourSettings` and moved both classes to `sc.fiji.ome.zarr.plugins.settings`

src/main/java/sc/fiji/ome/zarr/plugins/DnDHandlerPlugin.java
stefanhahmann: I moved the stuff from `sc.fiji.ome.zarr.settings` to `sc.fiji.ome.zarr.open.options`.
I chose `options` instead of `settings` to make it different to `sc.fiji.ome.zarr.plugins.settings`

src/main/java/sc/fiji/ome/zarr/settings/ZarrOpeningSettings.java
stefanhahmann: I moved the stuff from `sc.fiji.ome.zarr.settings` to `sc.fiji.ome.zarr.open.options`.
I chose `options` instead of `settings` to make it different to `sc.fiji.ome.zarr.plugins.settings`

src/main/java/sc/fiji/ome/zarr/open/ClipboardActions.java
stefanhahmann: Good comment. I agree. However, I would not like to touch this just now, since this PR is already massive. To stay consistent, it would maybe mean to not only rename the package, but perhaps also classes (e.g. `ZarrOpenActions` ~ `ZarrReadActions` or `openWithSettings` ~ `readWithSettings`

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
stefanhahmann: I renamed it MILLIS for improved clarity

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
stefanhahmann: The AI came up with `isAccessible`. Would that be more meaningful?

Regarding the other ideas it said:
  - canBeDownloaded — slightly misleading, HEAD doesn't download anything
  - canBeReached — misses the content-type filter (an HTML error page is "reachable")
  - isValidURI / isFunctionalURI — "valid" implies syntax checking, not network probing

I renamed to `isAccessible`

src/main/java/sc/fiji/ome/zarr/util/ZarrUtils.java
stefanhahmann: I guess not. `try-with` works for classes that implement `AutoCloseable`. But here, we have `disconnect()`, not `close()`.

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
stefanhahmann: I discussed with AI and it came up with an example, why `ensureTrailingSlash` is needed:

```
System.out.println( URI.create( "http://host/dataset/" ).resolve( "zarr.json" ) ); // gets http://host/dataset/zarr.json as expected
System.out.println( URI.create( "http://host/dataset" ).resolve( "/zarr.json" ) ); // gets http://host/zarr.json and NOT http://host/dataset/zarr.json as expected
```

I would thus keep the code here as it is.

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
stefanhahmann: All three Zarr related Utils classes have now been merged into one `ZarrUtils` class.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/n5/N5PyramidBackend.java
stefanhahmann: And since this is related I changed the similar pattern also in ZarrJavaPyramidBackend: https://github.com/BioImageTools/ome-zarr-fiji-java/pull/74/changes/70dfe15291638c6d106793f09d783b92eed22f24

src/main/java/sc/fiji/ome/zarr/util/ZarrLocations.java
stefanhahmann: There is only `ZarrUtils` left now.

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
xulman: many thanks :+1:

src/main/java/sc/fiji/ome/zarr/util/ZarrUtils.java
xulman: Aah, okay. Sorry, I overlooked.

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
xulman: okay, good counter example :)

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
xulman: VI suggests `isHttpAccessible` (or `isAccessibleViaHttp`) as that is exactly what it does: tests/checks if a given file is readable (w/o actually reading it) **via** HTTP protocol...

I know it is a silly detail. But, envision we could also have `isGoogleDriveAccessible()`, `isS3Accessible()`, `isOwnCloudAccessible()`... so let's get ready for it
(Yes, these are probably not DnD goals, but DnD I see as only one way of using this code, as only one client code out of many other different possible clients.)

only my two cents... :-)

xulman: Hmm... but the proposed future ones would also require providing credentials -> They will never show up in this *util* class, thus there's really no need to distinguish the communication layer when testing for a valid URL...

src/main/java/sc/fiji/ome/zarr/util/ZarrLocations.java
xulman: :+1:  many thanks for that

src/main/java/sc/fiji/ome/zarr/plugins/DnDHandlerPlugin.java
xulman: Good choice, btw. I really like this move. Many thanks for it.

src/main/java/sc/fiji/ome/zarr/open/ClipboardActions.java
xulman: Okay, it indeed would be too many things changed now. You're right, let's wait when the time is right.

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
xulman: Checked, nicely done. Many thanks again!

src/main/java/sc/fiji/ome/zarr/util/ClipboardUtils.java
```diff
@@ -0,0 +1,163 @@
...
+	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
+	 *   <li>{@code file:} URIs &ndash; used as-is</li>
+	 *   <li>plain filesystem paths &ndash; converted with
+	 *       {@link Paths#get(String, String...)}{@code .toUri()}</li>
+	 * </ul>
+	 * Reports a user-facing error and returns {@code null} when the clipboard
+	 * is empty, the text is not a recognizable URI/path, or the location does
+	 * not appear to be an OME-Zarr dataset.
+	 */
+	public static URI parseClipboardUri( final Consumer< String > errorHandler )
```
xulman: sorry, wouldn't `readClipboardAsUri()` be better?

src/main/java/sc/fiji/ome/zarr/util/ClipboardUtils.java
```diff
@@ -0,0 +1,163 @@
...
+	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
+	 *   <li>{@code file:} URIs &ndash; used as-is</li>
+	 *   <li>plain filesystem paths &ndash; converted with
+	 *       {@link Paths#get(String, String...)}{@code .toUri()}</li>
+	 * </ul>
+	 * Reports a user-facing error and returns {@code null} when the clipboard
+	 * is empty, the text is not a recognizable URI/path, or the location does
+	 * not appear to be an OME-Zarr dataset.
+	 */
+	static URI parseClipboardUri( final String clipboard, final Consumer< String > errorHandler )
```
xulman: Here, for instance, I thought that the parameter `clipboard` is actually a reference to some system clipboard, when, from the code later, I learned that this is actually the clipboard content that's only to be converted to URI.

I mean, well, looks again as a true product of AI and it's tendency to over-separate tasks:
This function is a fore-runner that checks its parameter for sanity to pass it further to (`toURI`) a proper String->URL converter... uff.. .which `triesParseURI` even outside itself...  and so the `toURI` is basically only a post-cleaner/post-checker after the conversion.

A very soft request for a short spaghetti: Could we handle the String to URI simply in one function?

src/main/java/sc/fiji/ome/zarr/util/ClipboardUtils.java
xulman: `readClipboard()` -> `String` is great, could be even public in my opinion
`readClipboardAsURI()` -> `URI` is great, already public

and then the `parseClipboardUri(str, errH)` + `tryParseUri()` + `toUri(..)` I would melt into one method that could also be public or package-protected at least :)

xulman (dismissed):
magnificent! 
many thanks!
We could do HTTPS now, yay! 💯

src/main/java/sc/fiji/ome/zarr/util/ClipboardUtils.java
xulman: ....into one method `URI stringToUri(String possibleUri, ErrorHandler)` or alike, with a lot of exceptions internally to trigger the ErrorHandler, and perhaps one Excpetion that goes outside of the metod (or just returning null) simply to flag somehow that the conversion failed.

xulman:
test passing on my Linux

src/main/java/sc/fiji/ome/zarr/util/ZarrOverHttpUtils.java
stefanhahmann: I renamed it to `isHttpAccessible`

src/main/java/sc/fiji/ome/zarr/util/ClipboardUtils.java
stefanhahmann: > sorry, wouldn't `readClipboardAsUri()` be better?

I renamed it.

src/main/java/sc/fiji/ome/zarr/util/ClipboardUtils.java
stefanhahmann: I have merged `readClipboardAsUri(String,…)`, `toUri()`, and `tryParseUri()` into a single public `stringToUri(String, Consumer<String>)`.
 
 I also made `readClipboard()` public.

PR #76, opened 2026-05-13
Simplify Pyramidal5DImageData open API

stefanhahmann:
## Summary

- Replace `(Context, URI, Integer preferredWidth)` factory-method signatures on `Pyramidal5DImageData` with `(Context, URI)` (no-settings) and `(Context, URI, ZarrOpeningSettings)` overloads, removing the raw `Integer preferredWidth` parameter from the public API
- Add `ZarrOpeningSettings.defaultSettings()` factory method and `effectivePreferredWidth()` helper, which encapsulate the "null width for highest resolution" logic that was previously scattered across callers
- Add `BackendMismatchException` (unchecked) thrown when `openWithN5`/`openWithZarrJava` receive settings that specify the wrong backend, making misuse of the specific-backend methods fail fast with a clear message
- Simplify `ZarrOpenActions.openMultiScaleImage()` to delegate entirely to `Pyramidal5DImageData.open()`, removing the inline switch on backend
- Remove the `PrefService` look-up from `Pyramidal5DImageData.open()`; settings are now passed in by the caller rather than read from persisted preferences inside the factory method
- Make `ZarrOpenActions` constructors package-private (they were only ever called from within the package)

## Test plan

- [ ] `mvn test` passes (all existing tests updated to new API, new `BackendMismatchException` tests added)
- [ ] `testOpenWithN5ThrowsOnZarrJavaBackendSettings` – `openWithN5` rejects ZARR_JAVA settings
- [ ] `testOpenWithZarrJavaThrowsOnN5BackendSettings` – `openWithZarrJava` rejects N5 settings
- [ ] `testOpenWithDefaultSettingsUsesN5Backend` – no-arg `open()` uses N5 as default
- [ ] `testOpenWithPreferredWidthN5` / `testOpenWithPreferredWidthZarrJava` – resolution selection still works via settings

### DIFF
--- a/src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
+++ b/src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
@@ -55,16 +55,14 @@ import net.imglib2.util.Cast;
 
 import bdv.util.BdvFunctions;
 import ij.IJ;
-import sc.fiji.ome.zarr.pyramid.Pyramidal5DImageDataImpl;
+import sc.fiji.ome.zarr.pyramid.Pyramidal5DImageData;
 import sc.fiji.ome.zarr.pyramid.exceptions.MultiImageDatasetException;
 import sc.fiji.ome.zarr.pyramid.exceptions.NoMatchingResolutionException;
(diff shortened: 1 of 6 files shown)


stefanhahmann:
Closed, since it became obsolete due to API changes.

PR #80, opened 2026-05-21
Add "open selected resolution level" as dataset functionality

### LINKED ISSUE #14: Indicate to user that an image is multi-resolution
I'm thinking it could be useful for a user to know that an image is multi-resolution, such that they know that they could go to a menu entry (e.g. `Image > Resolutions....`) to select another resolution. 

What about adding this to the image title similar to the `(V)` to indicate that it's a virtual stack?

So, for example, if people open an OME-Zarr, it could look like that: `my image (V,R)`

### LINKED ISSUE #41: Extend `PyramidalDataset` to get its underlying `Pyramidal5DImageData`
...this basically means to add

```java
Pyramidal5DImageData getPyramidal5DImage() {
    return this.data;
}
```

into the `PyramidalDataset`.

That way, Jython (and Groovy and other Java-based) scripts can easily cast to this more powerful type and benefit from its features.

### LINKED ISSUE #79: Implement dataset lifecycle for multiresolution OME-ZARRs in FIJI
From a discussion with @tpietzsch and @xulman we had the following idea how to handle multiresolution OME-Zarr datasets in FIJI
 
- Pyramidal5DImageData
  - not a dataset itself
  - has reference to images / zarr (volatile and non volatile), up to 5D (xyzct)

- For each resolution level that you open as a FIJI window
  - a dataset is created 
    - this dataset has a reference to the same Pyramidal5DImageData object
    - if the same resolution is opened multiple times, this results in multiple datasets all pointing to the same Pyramidal5DImageData object
    - each dataset is represented by its FIJI / ImageJ window
- If you open any of the opened resolutions in BigDataViewer
  - a new dataset is created
  - this dataset is "invisible" (it is only visible in the BDV window, but not in a FIJI / ImageJ window)
  - this dataset represents the BDV Window
  - this dataset has a reference to Pyramidal5DImageData
  - the BigDataViewer can use all resolution levels to progressively load image data

Scenario 1:
- dataset with 5 resolutions: open all resolutions level in ImageJ
  - results in 5 datasets
- open any of them in BDV:
  - results in a 6th dataset
- open any of them in BDV again:
  - results in a 7th dataset

Scenario 2:
- dataset with 5 resolutions: open this in BigDataViewer
  - results in 1 dataset
- open resolution level 3 from it (e.g. via the menu)
  - results in 2 datasets

stefanhahmann:
Resolves https://github.com/BioImageTools/ome-zarr-fiji-java/issues/41
Resolves https://github.com/BioImageTools/ome-zarr-fiji-java/issues/79
Resolves https://github.com/BioImageTools/ome-zarr-fiji-java/issues/14

## Summary

- Adds `OpenResolutionLevelCommand` (menu: *Plugins > OME-Zarr > Open Resolution Level…*) to open a specific resolution level of an OME-Zarr multi-resolution dataset as a new ImageJ dataset
- Adds per-level dataset access (`asPyramidalDataset(int)`) and `NonExistingResolutionLevelException` in the pyramid API
- Replaces `BdvHandleService` with `BdvFocusService`, which tracks the most recently focused BDV window via `WindowFocusListener`; commands inject `BdvFocusService` and call `resolveDataset()` to fall back to the focused BDV dataset when the standard `@Parameter Dataset` injection returns null (BDV windows are plain Swing frames, not SciJava `ImageDisplay`s)
- `OpenInBDVCommand` creates a distinct `PyramidalDataset` for each BDV open, sharing the same underlying `Pyramidal5DImageData` (pixel data loaded only once)
- Stores axis calibration per resolution level in `PyramidContents` / `AxisCalibration`; physical extent (pixels × scale) is verified to be identical across all resolution levels
- Selecting resolution level via the menu is not yet part of this PR to not make it too big.
- Showing dimensions (X*Y*Z*C*T) during the resolution level selection is also not yet part of this PR to not make it too big.

## Test plan

- [ ] Run `mvn test` — all existing and new tests pass

**Open in IJ → Open Resolution Level**
- [ ] Drag-and-drop an OME-Zarr onto Fiji → opens as IJ2 dataset (1 dataset in DatasetService)
- [ ] Run *Plugins > OME-Zarr > Open Resolution Level…* with the IJ2 dataset active → dialog shows resolution choices, selecting a level opens a new IJ2 window (2 datasets total); both share the same pyramid data

**Open in BDV → Open Resolution Level**
- [ ] Drag-and-drop an OME-Zarr onto Fiji → open in BDV (1 dataset in DatasetService)
- [ ] With the BDV window focused, run *Open Resolution Level…* without an active IJ2 display → command falls back to the focused BDV dataset, opens the selected level as a new IJ2 window (2 datasets total)

**Open in IJ → Open in BDV**
- [ ] Drag-and-drop → opens as IJ2 dataset (1 dataset)
- [ ] Run *Plugins > OME-Zarr > Open in BDV* with the IJ2 dataset active → BDV window opens (2 datasets total); BDV dataset is a distinct `PyramidalDataset` sharing the same underlying pyramid data as the IJ2 dataset

**Open in BDV → Open in BDV again**
- [ ] Drag-and-drop → open in BDV (1 dataset)
- [ ] With BDV window focused, run *Open in BDV* again (no active IJ2 display) → command falls back to focused BDV dataset, a second BDV window opens (2 datasets total); both BDV datasets share the same underlying pyramid data

**Open dataset A in IJ → Open dataset B in BDV → Open resolution level**
- [ ] Drag-and-drop → open in IJ (1 dataset)
- [ ] Drag-and-drop → open in BDV (2 datasets)
- [ ] With BDV window active run Open Resolution Level, select resolution level 1; The resolution level 1 of the dataset open in the BDV window should be shown.

**Error cases**
- [ ] Run *Open Resolution Level…* or *Open in BDV* with no image open → command cancels with a "No image is currently open" message dialog
- [ ] Close the BDV window, then run *Open in BDV* → same cancellation message


src/main/java/sc/fiji/ome/zarr/util/BdvUtils.java
```diff
@@ -79,21 +96,54 @@ public static BdvHandle showBdvAndRegisterDataset( final PyramidalDataset< ? > p
...
+
+	/**
+	 * Registers the new focused dataset with {@code bdvFocusService} and installs a
+	 * {@code WindowFocusListener} to keep the focus state up to date as the user
+	 * switches between BDV windows.
+	 */
+	private static void registerFocusTracking( final PyramidalDataset< ? > pyramidalDataset,
+			final Window window, final BdvFocusService bdvFocusService )
+	{
+		bdvFocusService.notifyWindowFocused( pyramidalDataset );
```
xulman: I would have liked to see this line between L121 and L122 -- that's right after `incrementReferences()` in the method `registerDatasetLifecycle()`. So that this method can really only deploy the async focus listener (while the `registerDatasetLifecycle()` is more two-fold: a sync/immediate register and async deregister).

src/main/java/sc/fiji/ome/zarr/util/BdvUtils.java
```diff
@@ -79,21 +96,54 @@ public static BdvHandle showBdvAndRegisterDataset( final PyramidalDataset< ? > p
...
 
+	/**
+	 * Notifies SciJava about the start of usage of this dataset and decrements the reference count
+	 * when the BDV window closes. Only called when we're able to listen for the window-close event.
+	 * If {@code bdvFocusService} is non-null, also notifies it when the window closes.
+	 */
+	private static void registerDatasetLifecycle( final PyramidalDataset< ? > pyramidalDataset,
+			final Window window, final BdvFocusService bdvFocusService )
+	{
+		pyramidalDataset.incrementReferences();
```
xulman, line 116: Biased from my C++ days, I still appreciate the pattern "something done here should be undone here as well", meaning the window closing listener *decrements* and *defocuses*, but the function *increments* but doesn't *focuses*. It could notify the focus as it is executed on the newly created window.

Together with the note below, it is a cosmetic detail after all.

src/main/java/sc/fiji/ome/zarr/plugins/OpenInBDVCommand.java
```diff
@@ -29,41 +29,49 @@
...
 	@Override
 	public void run()
 	{
-		logService.log( LogLevel.DEBUG, "Trying to open: " + dataset.getName() );
-		logService.log( LogLevel.DEBUG, "Class: " + dataset.getClass() );
-		logService.log( LogLevel.DEBUG, "Dataset instanceof PyramidalDataset: " + ( dataset instanceof PyramidalDataset ) );
-		if ( dataset instanceof PyramidalDataset )
+		if ( dataset == null && imageDisplayService != null )
+			dataset = imageDisplayService.getActiveDataset();
+		final Dataset resolved = bdvFocusService != null ? bdvFocusService.resolveDataset( dataset ) : dataset;
```
xulman, line 62: So, here a `Dataset` is provided (into the `resolved`) only if the `imageDisplayService` saw no active dataset. Is the `imageDisplayService` notified "not to see an active dataset" when BDV gained focus? (as the BDV focus gaining informs only the `BdvFocusService`, not the `ImageDisplayService`...)

xulman: suspecting I could open one dataset A in BDV window, and dataset B in IJ window, then don't I always get dataset B (as long as I don't close the IJ window) into `resolved`?

src/main/java/sc/fiji/ome/zarr/plugins/DnDHandlerPlugin.java
```diff
@@ -60,7 +60,7 @@ public class DnDHandlerPlugin extends AbstractIOPlugin< Object >
 	private static final Object FAKE_INPUT = new ArrayList<>( 0 );
 
 	@Parameter
-	private BdvHandleService bdvHandleService; //TODO, is it really used down-stream?
+	private BdvFocusService bdvHandleService; //TODO, is it really used down-stream?
```
xulman: Maybe we could really drop the reference here as the TODO is suggesting...

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
```diff
@@ -166,23 +170,131 @@ public void openViewerDialog()
...
-		return openImage(
-				BdvUtils::showBdvAndRegisterDataset,
-				singleScaleImage -> BdvFunctions.show( singleScaleImage, "Image" ),
-				"BigDataViewer"
-		);
+		try
+		{
+			return openPyramidImage(
+					() -> {
+						final PyramidalDataset< ? > dataset = Cast.unchecked( getPyramid().asPyramidalDataset() );
```
xulman: the other two functions above were not casting here

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
```diff
@@ -166,23 +170,131 @@ public void openViewerDialog()
...
+						final BdvFocusService bdvHandleService = context.getService( BdvFocusService.class );
```
xulman: old name of the variable

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
```diff
@@ -166,23 +170,131 @@ public void openViewerDialog()
...
+		catch ( NoMatchingResolutionException e )
+		{
+			showNonMatchingResolutionError( e );
+		}
+		return null;
+	}
+
+	private Pyramidal5DImageDataImpl< ?, ? > getPyramid()
+	{
+		if ( cachedPyramid == null )
```
xulman, line 247: Did I get it right that this is here only for lazy accessing the dataset around which this class is constructed?

Otherwise, one could have had `final` on the `cachedPyramid` attrib and called this method from c'tors...

xulman (dismissed):
I'm already submitting to see what's bugging my head, I'd like to study again in the IDE and test two things

- the (not?)notification case of the `imageDisplayService`
- is gain focus event arriving also for newly created windows

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
```diff
@@ -78,7 +78,7 @@ public final class PyramidContents<
 
 	public final RandomAccessibleInterval< V >[] volatileImgs;
 
-	public final AxisCalibration[] axes;
+	public final AxisCalibration[][] axesPerLevel;
```
xulman, line 84: Only a comment here that it may not necessarily be apparent if the first index is the resolution level and the second index is then the axis/dimension, or if the order is the opposite.

Biased by sitting between Python (numpy) and Java, where in the former, the order is usually reversed (to what a natural expectation would be).

---
If we wanna go already towards a more bulletproof representation of pyramidal content, maybe we want to have two-tier classes, e.g., featuring a method like this one: `AxisCalibration[] PyramidCalibration::getAxesCalibrationForLevel(int)`.

src/main/java/sc/fiji/ome/zarr/pyramid/exceptions/NonExistingResolutionLevelException.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class NonExistingResolutionLevelException extends RuntimeException
+{
+	public NonExistingResolutionLevelException( final int resolutionLevel, final int numResolutionLevels )
+	{
+		super( "Resolution level " + resolutionLevel + " does not exist. "
+				+ "The pyramid has " + numResolutionLevels + " level"
+				+ ( numResolutionLevels == 1 ? "" : "s" ) + " (0–" + ( numResolutionLevels - 1 ) + ")." );
+	}
+}
```
xulman: Just noticed: Wasn't the habit when introducing a new exception to also introduce a c'tor with (..., `Throwable cause`)? 

We are probably not in need of it now, I know.

src/main/java/sc/fiji/ome/zarr/util/BdvUtils.java
stefanhahmann: Good point. Moved `notifyWindowFocused` into `registerDatasetLifecycle()` right after `incrementReferences()`. Both are now treated symmetrically as immediate open-time setup (alongside the existing `notifyWindowClosed` in the close handler), and `registerFocusTracking()` is reduced to just deploying the async WindowFocusListener. Updated the Javadocs accordingly.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageData.java
```diff
@@ -64,13 +64,26 @@ public interface Pyramidal5DImageData< T extends NativeType< T > & RealType< T >
 	 */
 	PyramidalDataset< T > asPyramidalDataset();
 
+	/**
+	 * @param resolutionLevel 0-based index into the resolution pyramid (0 = highest resolution)
+	 * @return a fresh {@link PyramidalDataset} wrapping the image at the specified resolution level,
```
xulman: "A fresh" I propose to say "A newly created object of"

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageData.java
```diff
@@ -64,13 +64,26 @@ public interface Pyramidal5DImageData< T extends NativeType< T > & RealType< T >
...
+	 *   backed by the same underlying pyramid data
+	 */
+	PyramidalDataset< T > asPyramidalDataset( int resolutionLevel );
+
 	/**
 	 * @return an IJ2 {@code net.imagej.Dataset} wrapping the full-resolution
 	 *   5D (XYZCT) image; this will indirectly also serve the
 	 *   {@link net.imagej.ImgPlus}
 	 */
 	Dataset asDataset();
```
xulman, line 80: What resolution level does it open then? Could we have a default implementation?

Aah, now I read "full-resolution" in the docs. Hmm, I would also add "resolution level 0" or something similar that explicitly mentions the number zero '0', so it is clear I get here what I would have gotten by calling the *int*-parametrized sibling methods.

src/main/java/sc/fiji/ome/zarr/util/BdvUtils.java
stefanhahmann: Good point. Moved notifyWindowFocused into registerDatasetLifecycle() right after incrementReferences(). Both are now treated symmetrically as immediate open-time setup (alongside the existing notifyWindowClosed in the close handler), and registerFocusTracking() is reduced to just deploying the async WindowFocusListener. Updated the Javadocs accordingly.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
```diff
@@ -127,6 +128,10 @@ public class Pyramidal5DImageDataImpl<
 
 	private final Dataset ijDataset;
 
+	private final CachedCellImg< T, ?>[] cachedCellImgs;
+
+	private final AxisCalibration[][] axesPerLevel;
```
xulman, line 135: Please excuse me here as well for a similar comment as I did elsewhere about the idea of having a class that collects `AxisCalibration[]` for every resolution level.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
```diff
@@ -127,6 +128,10 @@ public class Pyramidal5DImageDataImpl<
...
+
 	private final List< SourceAndConverter< T > > sourceAndConverters;
```
xulman, line 138: Another observation regarding the indexing:

- `cachedCellImgs` and `sourceAndConverters` are indexed by number of channels, no?
- `axesPerLevel` are indexed first by resLevel and then by dimensions of the images

Less-careful users may be tempted to use the "channel index" also for the `axesPerLevel`... aka "accidents happen easily". It's probably their fault, but maybe we better API we could make this mistake less likely to happen.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
```diff
@@ -127,6 +128,10 @@ public class Pyramidal5DImageDataImpl<
 
 	private final Dataset ijDataset;
```
xulman, line 129: hmm... Do we need this one now when the images at a particular resolution are spawned on demand?

src/main/java/sc/fiji/ome/zarr/pyramid/PyramidalDataset.java
```diff
@@ -58,6 +59,22 @@ public PyramidalDataset( Pyramidal5DImageData< T > data )
 		super( data.asDataset().context(), data.asDataset().getImgPlus() );
 
 		this.data = data;
+		if ( data.numResolutionLevels() > 1 )
+			setName( multiResolutionName( data.getName() ) );
+	}
+
+	public PyramidalDataset( Pyramidal5DImageData< T > data, int resolutionLevel )
+	{
+		super( data.asDataset().context(), Cast.unchecked( data.asDataset( resolutionLevel ).getImgPlus() ) );
```
xulman: This Cast.unchecked() is not required.

src/main/java/sc/fiji/ome/zarr/plugins/DnDHandlerPlugin.java
stefanhahmann: Yes. I deleted it.

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
stefanhahmann: Removed the cast — good catch, thanks!

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
stefanhahmann: Renamed it.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageData.java
xulman: I have two small concerns here:

##### `asDataset` vs. `asPyramidalDataset`

Since the latter is an extension of the former, the methods that return the narrower type are not necessary at first -- but okay, let's make it explicit/obvious, but then I would have expected that `asDataset(3)` and `asPyramidalDataset(3)` will work exactly the same (except for returning different types) - and that's not what I'm seeing in the code.

In fact, the `asDataset` could have a default implementation, no?
e.g.

```java
default Dataset asDataset() {
		return asPyramidalDataset();
	}
```

##### use-case for the `asDataset()` -- w/o explicit res. level

I'm wondering if `asDataset()` could basically work as a duplication method? If I get a hold of some `Dataset` object and would like to create another one representing the same image *at the same resolution*, I don't have a good way of doing it. Instead, I can obtain a full-res image which is also easy to get with `asDataset(0)`.. especially easy since '0' refers to the full res, that makes the `asDataset()` not really special and worth having (unless, as stated above, we would make it work as a duplication method).

xulman: **Attention**:  The implementation of this method returns the current internal `ijDataset` (from the `...Impl` class) that is **at user-selected res. level**, not necessarily the full-resolution on as promised in the docs!

In fact, the current implementation works as the duplication method :)... still I would implement it by calling the `asPyramidalDataset()` (that would duplicate it) and narrowing the return type...

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
```diff
@@ -318,18 +326,47 @@ private SourceAndConverter< T > createSourceAndConverter( final RandomAccessible
...
+	}
+
 	@Override
 	public Dataset asDataset()
 	{
 		return ijDataset;
 	}
 
+	@Override
+	public Dataset asDataset( final int resolutionLevel )
```
xulman, line 357: This is reiterating the same comment again and again, I'm sorry, but couldn't we be creating here a proper `PyramidalDataset` returned back as the pure `Dataset`, or what's the motivation to strip down into a pure `Dataset` when this one is still backed by the shared pixels?

xulman (commented):
Since I found this rather an important (not a cosmetic lightweight) re-arrangement of the package, I tried to think about the changes a lot and from various perspectives.... and it unf. led to a lot of comments, some worries, and thus some suggestions...

src/main/java/sc/fiji/ome/zarr/plugins/OpenResolutionLevelCommand.java
```diff
@@ -0,0 +1,105 @@
...
+	public void initialize()
+	{
+		if ( bdvHandleService != null )
+			dataset = bdvHandleService.resolveDataset( dataset );
+		if ( dataset == null )
+		{
+			cancel( "No image is currently open." );
+			return;
+		}
+		if ( !( dataset instanceof PyramidalDataset ) )
```
xulman, line 76: Here, for example, we cannot function if the user has previously called `asDataset(int)` as this is creating a plain `Dataset`, not `PyramidalDataset`.  But maybe it was intended?  (I would be in favor of not restricting on the constructing/upfront end, but rather on the usage end, if restricting is really needed).

src/main/java/sc/fiji/ome/zarr/plugins/OpenResolutionLevelCommand.java
```diff
@@ -0,0 +1,105 @@
...
+	public void run()
+	{
+		if ( !( dataset instanceof PyramidalDataset ) )
+		{
+			logService.error( "Cannot open resolution level: the active image is not an OME-Zarr pyramidal dataset." );
+			return;
+		}
+		final PyramidalDataset< ? > pyramidalDataset = Cast.unchecked( dataset );
+		final int level = Integer.parseInt( resolutionLevel.replace( "Resolution ", "" ) );
+		final PyramidalDataset< ? > levelDataset = pyramidalDataset.getPyramidal5DImageData().asPyramidalDataset( level );
```
xulman, line 102: This is the dream come true.... was one of the goals of this entire project!!

Many thanks for it :-)

xulman (commented):
(forgotten part of the review)

(I haven't checked the tests yet too...)

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
stefanhahmann: Not only lazy access, also to share the same backing pyramid across multiple `openIJWithImage(int)` calls on the same instance.

At the moment is is hypthetical case, but when I think of developers using our API, it might be a possible case.

I acknowledged your hint and made it final and also initialized it in the constructor.

src/main/java/sc/fiji/ome/zarr/pyramid/PyramidalDataset.java
stefanhahmann: Removed it.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/PyramidContents.java
stefanhahmann: For now, I added some javadoc to clarify: `Axes per resolution level: {@code [resolutionLevel][axisIndex]}`.

src/main/java/sc/fiji/ome/zarr/pyramid/exceptions/NonExistingResolutionLevelException.java
stefanhahmann: We are inconsistent in our repo at the moment. I tend not to add this constructor, since I would argue, it is not needed for this type of Exception.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageData.java
stefanhahmann: There's a circular dependency. An implementation that doesn't override asDataset() at the moment would have this call chain:

``` 
  asDataset()                          // default: calls asPyramidalDataset()
    → asPyramidalDataset()             // impl: new PyramidalDataset<>(this)
      → PyramidalDataset(data)         // constructor: calls data.asDataset()
        → asDataset()                  // default again → stack overflow
```

I fixed the javadocs to document the user-selected resolution level.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
stefanhahmann: I would leave this for now.  The Javadoc addition on axesPerLevel (clarifying [resolutionLevel][axisIndex]) should address the readability concern you had.

src/main/java/sc/fiji/ome/zarr/pyramid/Pyramidal5DImageDataImpl.java
stefanhahmann: Actually `cachedCellImgs[]` is indexed by resolution level, not channel. Only `sourceAndConverters` is indexed by channel. But you're right that having them adjacent with different index semantics is confusing. I added javadoc to the fields to make this more explicit.

I would not make further changes to the API in this PR, if you agree?

src/main/java/sc/fiji/ome/zarr/util/BdvUtils.java
xulman: Many thanks!  Little detail, I know, but still appreciate it **very** much.

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
xulman: My bad, there was one more left .... but we resolved together.

src/main/java/sc/fiji/ome/zarr/open/ZarrOpenActions.java
stefanhahmann: I actually had to re-introduce this pattern, since the current exception handling currently relies on the Pyramidal5DImageDataset object not to be constructed already in the constructor.

src/main/java/sc/fiji/ome/zarr/plugins/OpenInBDVCommand.java
stefanhahmann: Thanks for pointing this out. This use case was not yet covered. Together with AI the code was rewritten to use AWT keyboard focus manager to track focus of all windows, which covers the use case you described.

stefanhahmann:
Closed in favor of #90

PR #84, opened 2026-05-27
Add support for opening OME-Zarr datasets in public S3 stores

stefanhahmann:
Enables opening `s3://bucket/path` URIs in both the ZarrJava and N5 backends.

**What's included:**
- Accept `s3://` URIs in `ClipboardUtils` and skip the `isZarr()` probe for them (which would require a short-lived S3 client just for detection)
- `ZarrJavaPyramidBackend`: builds an AWS SDK `S3Client` (default region `US_EAST_1`; standard AWS credential chain with an anonymous-credentials fallback) and constructs a zarr-java `S3Store` from the URI's host and path. The AWS SDK ships transitively with `zarr-java`.
- `N5PyramidBackend`: relies on `N5Factory`'s existing S3 support (via `n5-aws-s3`), only passing `Region.US_EAST_1` as the default region so environments without `~/.aws/config` still work. The `n5-aws-s3` dependency was added to the `ome-zarr-n5` module.

**Error handling:**
- Store-access failures (S3 auth failures, missing buckets, network errors, …) are RuntimeExceptions that previously escaped every catch block and surfaced as an unhandled exception in Fiji. Each backend now wraps them, in `load()`, into a backend-agnostic `StoreAccessException` (added to `ome-zarr-imglib2`): `ZarrJavaPyramidBackend` wraps zarr-java's `StoreException`, `N5PyramidBackend` wraps N5's `N5Exception` thrown while opening the reader / reading multiscale metadata.
- The Fiji opener (`ZarrOpener`) catches the single `StoreAccessException` and reports it as a user-facing error. This keeps the Fiji layer decoupled from the concrete backends.

**Authentication:**
* Works with public (anonymous) buckets and with private buckets whose credentials are available through the standard AWS chain (`~/.aws/credentials`, `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` env vars, or instance role credentials).

**Not yet included:**
* No UI to enter an S3 URI, bucket, region, or credentials. Users must paste an `s3://` URI via the existing clipboard/drag-and-drop path.
* There is currently no API to inject custom credentials programmatically.

* [x] Can be tested with this URI from the open organelle project: **s3://janelia-cosem-datasets/jrc_mus-choroid-plexus-3/jrc_mus-choroid-plexus-3.zarr/recon-1/em/fibsem-uint8**

### DIFF
--- a/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
+++ b/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
@@ -36,6 +36,7 @@ import java.awt.datatransfer.UnsupportedFlavorException;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
 import java.util.function.Consumer;
@@ -136,7 +137,7 @@ public final class ClipboardUtils
 		{
 			parsed = new URI( text );
 		}
-		catch ( Exception e )
+		catch ( URISyntaxException e )
 		{
 			logger.debug( "Text is not valid URI syntax, will try as a local path: {}", e.getMessage() );
 		}
--- a/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
+++ b/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
@@ -84,6 +84,7 @@ public final class ClipboardUtils
 	 * Handles three input forms:
 	 * <ul>
 	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
+	 *   <li>{@code s3://} URIs &ndash; used as-is</li>
 	 *   <li>{@code file:} URIs &ndash; used as-is</li>
 	 *   <li>plain filesystem paths &ndash; converted with
 	 *       {@link Paths#get(String, String...)}{@code .toUri()}</li>
@@ -99,9 +100,10 @@ public final class ClipboardUtils
 
 	/**
 	 * Converts a string to a {@link URI} suitable for opening an OME-Zarr dataset.
-	 * Handles three input forms:
+	 * Handles four input forms:
 	 * <ul>
 	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
+	 *   <li>{@code s3://} URIs &ndash; used as-is</li>
 	 *   <li>{@code file:} URIs &ndash; used as-is</li>
 	 *   <li>plain filesystem paths &ndash; converted with
 	 *       {@link Paths#get(String, String...)}{@code .toUri()}</li>
@@ -137,12 +139,13 @@ public final class ClipboardUtils
 		if ( parsed != null )
 		{
 			final String scheme = parsed.getScheme();
-			if ( "http".equalsIgnoreCase( scheme ) || "https".equalsIgnoreCase( scheme ) || "file".equalsIgnoreCase( scheme ) )
+			if ( "http".equalsIgnoreCase( scheme ) || "https".equalsIgnoreCase( scheme )
+					|| "file".equalsIgnoreCase( scheme ) || "s3".equalsIgnoreCase( scheme ) )
 				return parsed;
 			if ( scheme != null )
 			{
 				errorHandler.accept( "Unsupported URL scheme '" + scheme + "':\n" + text + "\n\n"
-						+ "Supported schemes are http, https, and file." );
+						+ "Supported schemes are http, https, file, and s3." );
 				return null;
 			}
 		}
--- a/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
+++ b/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
@@ -81,11 +81,14 @@ public final class ClipboardUtils
 
 	/**
 	 * Resolve text read from clipboard to a URI suitable for opening an OME-Zarr dataset.
-	 * Handles three input forms:
+	 * Handles four input forms:
 	 * <ul>
 	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
 	 *   <li>{@code s3://} URIs &ndash; used as-is</li>
-	 *   <li>{@code file:} URIs &ndash; used as-is</li>
+	 *   <li>{@code file:} URIs &ndash; used as-is; both {@code file:/path/to/data.zarr} and
+	 *       {@code file:///path/to/data.zarr} are usable. A URI naming a host
+	 *       ({@code file://some-server/path/to/data.zarr}) is rejected by the backends,
+	 *       because {@link Paths#get(URI)} accepts no authority component.</li>
 	 *   <li>plain filesystem paths &ndash; converted with
 	 *       {@link Paths#get(String, String...)}{@code .toUri()}</li>
 	 * </ul>
@@ -104,7 +107,10 @@ public final class ClipboardUtils
 	 * <ul>
 	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
 	 *   <li>{@code s3://} URIs &ndash; used as-is</li>
-	 *   <li>{@code file:} URIs &ndash; used as-is</li>
+	 *   <li>{@code file:} URIs &ndash; used as-is; both {@code file:/path/to/data.zarr} and
+	 *       {@code file:///path/to/data.zarr} are usable. A URI naming a host
+	 *       ({@code file://some-server/path/to/data.zarr}) is rejected by the backends,
+	 *       because {@link Paths#get(URI)} accepts no authority component.</li>
 	 *   <li>plain filesystem paths &ndash; converted with
 	 *       {@link Paths#get(String, String...)}{@code .toUri()}</li>
 	 * </ul>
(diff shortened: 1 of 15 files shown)


src/main/java/sc/fiji/ome/zarr/pyramid/backend/n5/N5PyramidBackend.java
```diff
@@ -101,7 +103,9 @@ public N5PyramidBackend( final URI inputUri )
 	@Override
 	public PyramidContents< T, V > load()
 	{
-		final N5Reader reader = new N5Factory().openReader( inputUri.toString() );
+		final N5Reader reader = new N5Factory()
+				.s3Configuration( builder -> builder.region( Region.US_EAST_1 ) )
```
xulman: I hope this is not too expensive, as it is used for *any* connection. Theoretically, it should remember the reference on a method to call it only when it needs it...

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
```diff
@@ -654,6 +658,67 @@ void testOpenSingleScaleImageInBDV( String resource ) throws URISyntaxException
...
+	{
+		try (Context context = new Context())
+		{
+			final AtomicReference< String > error = new AtomicReference<>();
+			final ZarrOpeningSettings settings = new ZarrOpeningSettings();
+			settings.setReaderBackend( backend );
+			final ZarrOpenActions actions = new ZarrOpenActions( S3_JANELIA_CHOROID_PLEXUS, context, settings, error::set );
+			final AtomicInteger multiScaleCounter = new AtomicInteger( 0 );
+			final AtomicInteger singleScaleCounter = new AtomicInteger( 0 );
+			actions.openImage( dataset -> multiScaleCounter.incrementAndGet(), img -> singleScaleCounter.incrementAndGet() );
```
xulman: This test function looks very similar to others... `testOpenValidMultiScaleImagePath()`, `testOpenValidSingleScaleImagePath()`,  maybe even close to `testOpenBioformats2rawCollectionRootReportsMultiImage()`.

Besides, the `actions.openImage()` is not used by any production code, it's called only from the tests... what's the purpose of unit testing it?  Is the purpose of `openImage()` only to catch one more exception than `openPyramidImage()` that has become now the main workhorse for opening?

xulman: One more comment w.r.t. `openImage()`. On the first sight, I couldn't understand why this test is of any use as the lambdas passed to the `openImage()` are only incrementing the counters, which is then tested; and I was asking myself where the ome-zarr loading is happening at all and how could it come that a counter would not be incremented... then I've noticed that the trick is in the caching inside `ZarrOpenActions`, that we're testing if the "cache filling" would fail or not. If we were do to anything about the `openImage()`, could we also think of a more obvious testing pattern (not sure if there's one)?

xulman (dismissed):
Nice work. Thanks.

One of the tests caught my attention, and I was "complaining" that perhaps we could reuse code and maybe even shorten/simplify the (testing) codebase?

Also, I'm thinking if it would be possible to make N5 and ZarrJava throw the *same* exception. Or is our layer swallowing them and re-sending further to a client code under one common exception?

stefanhahmann:
> Also, I'm thinking if it would be possible to make N5 and ZarrJava throw the _same_ exception. Or is our layer swallowing them and re-sending further to a client code under one common exception?

This is now the case. Both backends translate their native store-open failures into a single backend-agnostic exception, [StoreAccessException](https://github.com/BioImageTools/ome-zarr-fiji-java/blob/add-s3-support/ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/exceptions/StoreAccessException.java) (in the ome-zarr-imglib2 API module):                                                                                             
                                                                                                                                                                                                                                                                                                                   
  - ZarrJavaPyramidBackend wraps zarr-java's StoreException → [ZarrJavaPyramidBackend.java#L268](https://github.com/BioImageTools/ome-zarr-fiji-java/blob/3666ab3ec1b6a21882a85cd0b97b4ba7f75e6011/ome-zarr-zarrjava/src/main/java/ome/zarr/zarrjava/ZarrJavaPyramidBackend.java#L268)                                                                                                                                                                                                                     
  - N5PyramidBackend wraps N5's N5Exception → [N5PyramidBackend.java#L116 ](https://github.com/BioImageTools/ome-zarr-fiji-java/blob/3666ab3ec1b6a21882a85cd0b97b4ba7f75e6011/ome-zarr-n5/src/main/java/ome/zarr/n5/N5PyramidBackend.java#L116)                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                   
  The Fiji opener (ZarrOpener) then catches just that one type and surfaces it as a user-facing error.

src/main/java/sc/fiji/ome/zarr/pyramid/backend/n5/N5PyramidBackend.java
stefanhahmann: I have looked it up: the N5Factory.s3Configuration(...) is just a setter. It only stores the Consumer<S3ClientBuilder>                                                                                                                                                                                                                                                                    

To make that more explicit at our own layer, I've now guarded it behind an s3:// scheme check, so the region default is applied only for S3 locations: [N5PyramidBackend.java#L107-L109](https://github.com/BioImageTools/ome-zarr-fiji-java/blob/f79bbb4c96fb6b55250b246bcb67a94a5b8789df/ome-zarr-n5/src/main/java/ome/zarr/n5/N5PyramidBackend.java#L108).

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
stefanhahmann: I removed openImage() entirely from production code, since it is a bad pattern to have methods in production code that are used only in unit tests.

The purpose of openImage() was to run the full open pipeline headlessly: instead of showing the result in a UI (as openIJWithImage()/openBDVWithImage() would do), it applied caller-supplied functions to the opened dataset. This allowed having tests that would perform the whole load → cache → exception-routing path and observe the outcome without instantiating a UI window (having too many of these windows caused memory problems in the CI pipeline in the past).

The extra catch (NoMatchingResolutionException) was not the reason it existed.

Success tests now load headlessly via ZarrOpener.getContents() (no UI window); rejection tests go through the real openIJWithImage(), which never reaches the UI show on failure paths, so they still exercise the production error handling.

About the similar tests you flagged: the refactor removed most of the repetitive code. The counter stuff is gone. Success cases are now one liners with loadMultiscaleHeadless(...) helper calls. Rejection cases are a uniform openIJWithImage() + error-handler check. I kept all tests since each test covers a distinct input/outcome.

ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
```diff
@@ -84,6 +84,7 @@ public static String readClipboard()
 	 * Handles three input forms:
 	 * <ul>
 	 *   <li>{@code http://} or {@code https://} URLs &ndash; used as-is</li>
+	 *   <li>{@code s3://} URIs &ndash; used as-is</li>
 	 *   <li>{@code file:} URIs &ndash; used as-is</li>
```
xulman: cosmetic detail, this is maybe spelled also as `file://`... (applies, if at all, in other places too in this src file)

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
```diff
@@ -659,6 +657,64 @@ void testOpenSingleScaleImageInBDV( String resource ) throws URISyntaxException
...
+		try (Context context = new Context())
+		{
+			final URI uri = URI.create( "s3://nonexistent-bucket/some/path" );
+			final AtomicReference< String > capturedError = new AtomicReference<>();
+			final ZarrOpeningSettings settings = new ZarrOpeningSettings();
+			settings.setReaderBackend( backend );
+
+			if ( backend == ZarrReaderBackend.ZARR_JAVA )
+			{
+				try (MockedConstruction< ZarrJavaPyramidBackend > mock = mockConstruction(
```
xulman: I am no good on the Mocked* things, but why is it here for the `ZarrReaderBackend` and it is **not** used for the `N5Backend` (in N5 we go directly for the trying to open with `actions.openIJWithImage();`

it is probably my oversight, and not knowing

xulman (dismissed):

ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/util/ClipboardUtils.java
stefanhahmann: Kept `file:` deliberately — the check in the code is on the scheme only.
So: 

- `file:/path/to/data.zarr` works just as well as 
- `file:///path/to/data.zarr`

I expanded the Javadoc to show both spellings, and to note that a host / server component (`file://some-server/path`) is rejected by the backends since Paths.get(URI) accepts no authority.

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
stefanhahmann: Good catch. The asymmetry wasn't a mocking preference, it was hiding a real bug. 

Unlike N5, the zarr-java path never actually produced a `StoreAccessException`: S3Store reported S3 errors as the AWS SDK's unchecked `SdkException`, which slipped past the `StoreException`-only catch and reached the user as an unhandled 403 stack trace.

The `MockedConstruction` was covering that

Fixed this.

The mock could be removed after the fix.

Issue #81, opened 2026-05-21
Split repo into five focused artifacts


stefanhahmann:
## Motivation

The current monorepo bundles everything — imglib2 model, Fiji/BDV integration, backend I/O (N5, zarr-java), and UI — into a single artifact. This makes it impossible for non-Fiji Java tools (QuPath, Icy, custom pipelines) to depend on just the OME-Zarr reading logic without pulling in the full Fiji stack. It also forces Fiji plugin authors to take both backends even if they only need one.

## Proposed artifacts and dependency hierarchy

```
                      ome-zarr-imglib2  (API, no backend)
                     /        |        \
          ome-zarr-n5   ome-zarr-zarr-java   ome-zarr-fiji  (API, no backend)
                     \        |        /
                       ome-zarr-fiji-ui
```

**Both `ome-zarr-imglib2` and `ome-zarr-fiji` are API-only artifacts** (analogous to `slf4j-api`): they define interfaces and integration layers but ship no backend. Neither is useful in isolation — a backend must be present on the classpath. The backends (`ome-zarr-n5`, `ome-zarr-zarr-java`) pull in `ome-zarr-imglib2` as a transitive dependency, so downstream consumers rarely need to declare it directly.

### `ome-zarr-imglib2`  _(API, no backend)_
- No dependencies on Fiji, ImageJ, BDV, N5, or zarr-java
- Conatins: `Pyramidal` interface
- Contains: `PyramidalContents` class (backend-agnostic methods only: `numChannels`, `numTimepoints`, `numResolutionLevels`, `getType`, `getName`, `getOmeroProperties`), OME-Zarr metadata model, imglib2 types, coordinate transforms, channel/timepoint metadata
- Contains: `PyramidBackend` interface 
- Perhaps with a new `asImg()` method
- Non-Fiji developers (QuPath, Icy, …) should depend on `ome-zarr-n5` or `ome-zarr-zarr-java` directly; they get `ome-zarr-imglib2` transitively

### `ome-zarr-n5`
- Depends on `ome-zarr-imglib2` + N5-universe
- Provides a concrete N5-backed implementation of `PyramidBackend`
- **Fully usable standalone** for non-Fiji developers — adding this artifact gives complete OME-Zarr reading capability with no Fiji dependency

### `ome-zarr-zarr-java`
- Depends on `ome-zarr-imglib2` + zarr-java
- Same as above, backed by zarr-java instead

[...]

PR #93, opened 2026-06-18
Prepare packages for splitting into modular artifacts

stefanhahmann:
Prepares the codebase for the artifact split described in #81. The model, backends, Fiji integration and UI now live in cleanly separated packages that each map to a single future artifact, but the repo is *not yet actually split*.

The decision on *how* to split is still open: a single Maven multi-module repo (if all 5 artifacts can be produced from it) vs. multiple repos.

## Preparation TODOs from #81 addressed

- **`Pyramidal5DImageData`** — interface dissolved; `PyramidBackend` is now the backend interface returning `PyramidContents`; `load` is generic in the pixel type; backends instantiate independently of the URI; `asImg()` added.
- **`PyramidContents` cleanup** — reduced overspecification, name-based axis lookup (no more presence/index fields), derived `channelLabels` from Omero data, documented dimension count relations and coordinate tansforms, dropped `mpicbg…VoxelDimensions`.
- **Volatile (V) type** — `V` generic and `volatileImgs` removed from the imglib2 layer; `VolatileViews` wrapping moved to the Fiji layer (`PyramidalBdv`).
- **In-code TODOs** — `Pyramidal` stripped to `getPyramidContents()`; `preferredMaxWidth` handling moved out of `PyramidalDataset`; dataset decoupled from backend.
- **Layer hygiene** — packages reorganized so the model/backends carry no Fiji deps; `BackendBenchmark`/`BdvHandleService` moved to demo.

## Target package → artifact mapping

| Future artifact / layer | Packages | Responsibilities |
|---|---|---|
| **ome-zarr-imglib2** (API + core, imglib2-only) | `pyramid.backend`, `pyramid.metadata`, `pyramid.exceptions` | `PyramidBackend` backend interface, immutable `PyramidContents` (+`asImg()`), axis/omero metadata, affine helpers |
| **ome-zarr-n5** (backend) | `pyramid.backend.n5` | `N5PyramidBackend` — read via N5 universe |
| **ome-zarr-zarr-java** (backend) | `pyramid.backend.zarrjava` | `ZarrJavaPyramidBackend`, `ZarrJavaCellLoader` — read via zarr-java |
| **ome-zarr-fiji** (Fiji/BDV integration) | `pyramid.fiji`, `pyramid.fiji.util`, `pyramid.fiji.open` | `PyramidalDataset`/`PyramidalBdv`/`PyramidalService`, `BdvUtils`, `ZarrOpener` + backend selection |
| **ome-zarr-fiji-ui** (UI/commands) | `open`, `open.options`, `ui`, `plugins`, `util` | DnD handler, dialogs, SciJava commands, opening-behavior settings, clipboard, script, zarr-detection helpers |

## Note for the future

The non-Fiji artifacts (`ome-zarr-imglib2`, `ome-zarr-n5`, `ome-zarr-zarr-java`) still live under `sc.fiji.ome.zarr.*`. Renaming their package root so they no longer carry `sc.fiji` is an option worth considering as part of the split.


xulman (dismissed):
Very nice!

I haven't studied all details. Instead, I looked merely on the new layout (packages) and content/functionalities of their classes. It is looking good.

Minor idea would be:
For `fiji-ui` case, the packages structure explicitly groups into `open`, `utils`, `plugins`, etc. But non-ui `fiji` case seems to be following this, has also `open` and `utils`, but it is not having `plugins` while it has plugin classes.

I was also thinking of creating a top-level package like `sc.fiji.ome.zarr.fiji-ui.*`, then we would have top-level packages `pyramid` (for pure backend, for fiji "backend") and `fiji-ui` (for fiji front-end). Currently, when enabling artifacts in a project, their content is "mixing" in the `sc.fiji.ome.zarr` package.

Thinking also of Jython programmers who are discovering our classes :)
But all that is very minor.

stefanhahmann:
> Very nice!
> 
> I haven't studied all details. Instead, I looked merely on the new layout (packages) and content/functionalities of their classes. It is looking good.
> 
> Minor idea would be: For `fiji-ui` case, the packages structure explicitly groups into `open`, `utils`, `plugins`, etc. But non-ui `fiji` case seems to be following this, has also `open` and `utils`, but it is not having `plugins` while it has plugin classes.
> 
> I was also thinking of creating a top-level package like `sc.fiji.ome.zarr.fiji-ui.*`, then we would have top-level packages `pyramid` (for pure backend, for fiji "backend") and `fiji-ui` (for fiji front-end). Currently, when enabling artifacts in a project, their content is "mixing" in the `sc.fiji.ome.zarr` package.
> 
> Thinking also of Jython programmers who are discovering our classes :) But all that is very minor.

Thanks for the review! I addressed the first point: moved `PyramidalService` and `PyramidalPreprocessor` into a new `pyramid.fiji.plugins` subpackage, so the @Plugin classes are grouped their similarly to the same package in the fiji-ui side.                                                                                            
                                                                                                                                                                                                                                                                                                                   
On the top-level fiji-ui package idea: most of what would land there (open, plugins, util) is actually usable headless/from scripts, so wrapping it under a name that says "ui" would feel misleading to me. Currently, only the ui package itself is genuinely GUI-only.

The non-fiji related classes, I would like to move anyway into a package that does not carry the name "fiji" in it.

xulman:
> The non-fiji related classes, I would like to move anyway into a package that does not carry the name "fiji" in it.

Aah, true! That solves it fo rme :-) Many thanks for reminding me.

PR #99, opened 2026-08-07
Open a single OME-Zarr resolution level

stefanhahmann:
Adds support for opening an individual OME-Zarr resolution level (a dropped `.../0` array folder), replacing the old dead `N5Utils.open()` path.

A single array is modeled as a **one-level pyramid**, so it flows through the existing `PyramidalDataset` / `PyramidalBdv` pipeline unchanged. The chosen reader backend is honoured, and `ome-zarr-fiji` no longer depends on the raw N5 libraries.

**Dependencies**

Dropping `N5Utils.open()` was the last direct N5 usage in `ome-zarr-fiji`, so the module is now free of any N5 dependency (only test scope remains). Backend selection is therefore genuinely a `pom.xml` decision: shipping only `ome-zarr-zarrjava` no longer drags N5 onto the classpath.

**How a node is resolved**:

```mermaid
flowchart TD
    A["load(URI)"] --> B{"multiscales metadata?"}
    B -->|yes| C["pyramid — one level per resolution"]
    B -->|no| D{"parent group lists this array?"}
    D -->|yes| E["one-level pyramid<br/>axes + scale + OMERO from the parent<br/>(the only route for OME-Zarr v0.4 / Zarr v2 levels,<br/>which carry no metadata of their own)"]
    D -->|no| F{"array has dimension_names?"}
    F -->|yes| G["one-level pyramid, uncalibrated<br/>(Zarr v3 only)"]
    F -->|no| H["SingleArrayAxesUnknownException<br/>— refuse rather than guess"]
```

The parent is read lazily, only for the single-array case (≤ 1 small metadata request on remote stores).

**Notes**
- `PyramidBackend` gains no new method; `load(URI)` is broadened to accept array nodes. The ordering above lives once in a new `AbstractPyramidBackend`, which leaves only `loadMultiscale` / `tryLoadLevelFromParent` / `tryLoadArrayNodeOnly` to each backend.
- `PyramidContents` / `Pyramidal*` are otherwise untouched (only an additive `singleLevel` factory).
- The previously `@Disabled` single-scale tests are enabled and updated to the new behaviour; the `openImage` tests moved from `ome-zarr-fiji-ui` down to `ome-zarr-fiji`.
- `imglib2-ij` change to test scope in `ome-zarr-fiji` (unused by main sources), and `openIJWithImage()` is now `void` (it returned `null` in all cases).

**Preferred width: ask instead of refuse**

A preferred width that no resolution level satisfiying it used to end in an error dialog with a single OK button, so the only way forward was the settings panel. The coarsest level is now offered instead, with a choice:

grafik

- The check applies only where a single level is picked for an ImageJ window. BigDataViewer shows the whole pyramid and streams it lazily, and `openIJWithImage(int)` is told which level to use — neither consults the preferred width any more.
- A dropped single level that is too wide gets the same message, minus any claim about other levels


ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/ZarrUtils.java
```diff
@@ -149,6 +149,78 @@ private static URI ensureTrailingSlash( final URI uri )
...
+		if ( datasetPath.equals( lastSegment( childUri ) ) )
+			return true;
+		final String stripped = stripTrailingSlashes( childUri.toString() );
+		return stripped.endsWith( "/" + stripTrailingSlashes( datasetPath ) );
+	}
+
+	private static String stripTrailingSlashes( final String s )
+	{
+		int end = s.length();
+		while ( end > 0 && s.charAt( end - 1 ) == '/' )
```
xulman: also VI is noticing this is no good
at least check for non-`/` at `end-2`... still don't think it is a good idea

can't we convert to `URI` or `URL` even and use its (theoretically bullet-proof) functions....
👉 maybe the creation of URI itself would already strip away anything unimportant trailing

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/AbstractPyramidBackend.java
```diff
@@ -0,0 +1,158 @@
...
+	 * @return the level as a one-level pyramid, or {@code null} when the parent is
+	 *   not a readable multiscales group or does not list this array
+	 */
+	protected abstract < T extends NativeType< T > & RealType< T > > PyramidContents< T > tryLoadLevelFromParent(
+			URI parentUri, URI arrayUri );
+
+	/**
+	 * Attempts to open {@code arrayUri} purely from its own metadata, without a
+	 * parent multiscales group: typically from the Zarr v3
+	 * {@code dimension_names}, opened uncalibrated (unit scale, no units, no
```
xulman: "uncalibrated" is a little scary to me...

I'm thinking better to let go and refuse to open, rather than opening "halfways" (I don't care about OMERO ext., I'm worried about the scale and units.)

I have to say I don't understand fully (with which zarr versions, etc.) when this situation occurs. Is it a case for/because of the old data (zarr <=2, ngff <= 0.4)? If so, I would not mind to drop it and live only with the `tryLoadLevelFromParent()`

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/metadata/AxisCalibration.java
```diff
@@ -67,4 +67,25 @@ public AxisCalibration( final String name, final String unit, final double scale
...
+	 * <p>
+	 * The names must already be in the imglib2 F-order expected by
+	 * {@link ome.zarr.imglib2.PyramidContents}; a reader whose library hands out
+	 * Zarr's {@code dimension_names} in storage (C) order has to reverse them
+	 * first, just as it does for the shape.
+	 *
+	 * @param axisNames axis names in imglib2 F-order
+	 * @return axis calibrations in the same order, each with unit scale and no unit
+	 */
+	public static AxisCalibration[] fromAxisNames( final String[] axisNames )
```
xulman: Uff... I'm not excited about having such functionality overhere at all -- please, see also the comment about `AbstractPyramidBackend::tryLoadArrayNodeOnly()`.

Besides that, the name of the method doesn't say "uncalibrated". It easily makes me believe this is just another way of setting up the calibrated axes. And it it really needs to be here, couldn't we allow clients to decide their own "uncalibrated values", with default being `"",1.0`, or maybe even `"undefined",1.0` or something else more obvious?

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/AbstractPyramidBackend.java
xulman: Aha! Found in the upstream caller (`loadSingleArray()`) that this one is indeed mainly for zarr <=2, ngff <= 0.4
Hmm.... 🤔 :-P

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/PyramidContents.java
```diff
@@ -255,33 +255,75 @@ private int sizeAlongAxis( final String axisName )
...
+	/**
+	 * Returns the index of the highest-resolution level that is still no wider
+	 * than {@code preferredMaxWidth}, or 0 when {@code preferredMaxWidth} is
+	 * {@code null}.
+	 * <p>
+	 * When no level is narrow enough — including the trivial case of a
+	 * single-level pyramid that is already too wide — the coarsest level
+	 * ({@code numResolutionLevels() - 1}) is returned as the closest available
+	 * match. Callers for which an oversized image is a problem should therefore
+	 * compare the {@link #widthAtLevel} of the returned level against their own
```
xulman: That means that callers are expected to write more lines of code anyway (to obtain an image from unknown resolution level, and to check ("do the math themselves") if there is any reasonable res. level at all). I would have switched to API:

- `int suggestResolutionLevel( preferredMaxWidth )`, which can flag somehow that such level isn't available
- `Img<T> asSmallestImg()`
- rename or alias `asImg()` to `asLargestImg()`
- and we have already `asImg( int level )`

The pattern is then clearer, I think:
```
int goodLevel = pc.suggestResolutionLevel( preferences );
if (goodLevel == -1) return pc.asSmallestImg();
return pc.asImg(goodLevel);
```

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/PyramidContents.java
```diff
@@ -255,33 +255,75 @@ private int sizeAlongAxis( final String axisName )
...
-	 * {@code preferredMaxWidth} is {@code null}.
+	 * Width of the image at the given resolution level, i.e. its extent along x
+	 * (index 0 in imglib2 F-order).
 	 *
-	 * @throws NoMatchingResolutionException if {@code preferredMaxWidth} is
-	 *   smaller than the width of every resolution level
+	 * @throws IndexOutOfBoundsException if {@code resolutionLevel} is not in
+	 *   {@code [0, numResolutionLevels())}
+	 */
+	public long widthAtLevel( final int resolutionLevel )
```
xulman: Why here we know that x-axis is at 0, while in `sizeAlongAxis()` we have to obtain `axisIndex()` ??

I would drop this method, and prepared a two parameter `sizeAlongAxis(axisName, resLevel)` instead.
And let `sizeAlongAxis(axisName) { return sizeAlongAxis(axisName,0); }`

xulman (changes requested):
This is 1/N part processed. I can't anymore today, long day. To be continued ASAP.

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/ZarrUtils.java
```diff
@@ -149,6 +149,78 @@ private static URI ensureTrailingSlash( final URI uri )
...
+	{
+		if ( uri == null )
+			return null;
+		final String stripped = stripTrailingSlashes( uri.toString() );
+		final int slash = stripped.lastIndexOf( '/' );
+		if ( slash < 0 )
+			return null;
+		// Keep the trailing slash so the result is treated as a folder location.
+		return URI.create( stripped.substring( 0, slash + 1 ) );
+	}
```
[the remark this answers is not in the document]
xulman: I think a nicer parent function coudl look like this:

```java
	public static URI getParentURI( final URI uri )
	{
		// URIs are specifically formatted, made of various (and often optional) sections,
		// and only the path section can be traversed up (reducing number of '/')
		Path parentPath = Paths.get( uri.getPath() ).getParent();
		if ( parentPath == null )
		{
			// nowhere to go up, signal it
			return null;
		}
		if ( parentPath.isAbsolute() )
		{
			return uri.resolve( parentPath.toString() );
		}
		else
		{
			// the input 'uri' is (most likely) a relative local path,
			// let's keep it a local path (despite this is not what URIs ideally should be looking like)
			return URI.create( parentPath.toString() );
		}
	}
```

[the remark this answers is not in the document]
xulman: perhaps in conjuction with some like

```java
	public static URI getURI( final String uri )
	{
		return URI.create( uri ).normalize();
	}
```

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/ZarrUtils.java
xulman: `URI#normalize()` goes in that direction... removes cumulated slashes, yet works only within the *path* part of an URI/URL

ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/open/ZarrOpenActions.java
```diff
@@ -212,9 +211,9 @@ public void openViewerDialog()
 	 * Opens the dataset in ImageJ at the resolution selected by the settings.
 	 * Delegates to {@link ZarrOpener#openIJWithImage()}.
 	 */
-	public Object openIJWithImage()
+	public void openIJWithImage()
 	{
-		return opener.openIJWithImage();
+		opener.openIJWithImage();
```
xulman: good! many thanks for catching this

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
```diff
@@ -296,116 +276,16 @@ void openViewerDialogDoesNotThrow() throws URISyntaxException
...
+		Path path = ZarrTestUtils.resourcePath( "ome/zarr/testdata/5d_testing/5d_dataset_v4.ome.zarr" );
 		try (Context context = new Context())
 		{
-			Path path = ZarrTestUtils.resourcePath( "ome/zarr/testdata/5d_testing/5d_dataset_v4.ome.zarr" );
-			ZarrOpeningSettings settings = new ZarrOpeningSettings( ZarrOpenBehavior.IMAGEJ_CUSTOM_RESOLUTION, 10 );
-			ZarrOpenActions actions = new ZarrOpenActions( path.toUri(), context, settings, System.out::println );
-			assertDoesNotThrow( () -> {
-				actions.openIJWithImage();
-			} );
+			ZarrOpener opener = ZarrOpenActions.defaultOpener( path.toUri(), context );
```
xulman: Is this really testing whether the default *opener* has been used??

`ZarrOpenActions.defaultOpener()` is really only about taking the `ZarrOpeningSettings.DEFAULT_READER_BACKEND`, and nothing more

from the test function name, I was expecting this choice (of the backend) to be the subject of the test

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
```diff
@@ -477,90 +357,73 @@ void testOpenMultiScaleDatasetInImageJ( String resource ) throws URISyntaxExcept
...
+				assertArrayEquals( new long[] { 64, 64 }, dimensions );
 			}
 			if ( resource.contains( "3d_testing" ) )
 			{
 				if ( resource.contains( "xyc" ) )
 				{
-					assertArrayEquals( new int[] { 64, 64, 3, 1, 1 }, dimensions );
-					assertEquals( 3, channels );
-					assertEquals( 1, frames );
-					assertEquals( 1, slices );
```
xulman: **I'm giving up...**:  Why this test was okay (with 5 numbers) before and is no longer okay now and needs to be changed (to 3 numbers only)???  I totally don't get the context and motivation for the changes.

xulman (commented):
just another bit 2/N

I'm increasingly worried about the usage of AI, which is becoming more a happy patcher and gifted generator of LoCs....

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
```diff
@@ -477,90 +357,73 @@ void testOpenMultiScaleDatasetInImageJ( String resource ) throws URISyntaxExcept
...
+			actions.openIJWithImage();
+
+			DatasetService datasetService = context.getService( DatasetService.class );
+			assertNotNull( datasetService );
+			List< Dataset > datasets = datasetService.getDatasets();
+			assertNotNull( datasets );
+			// A single resolution level opens as a one-level PyramidalDataset, just like a multiscale image.
+			assertEquals( 1, datasets.size() );
+			Dataset dataset = datasets.get( 0 );
+			PyramidalDataset pyramidalDataset = Cast.unchecked( dataset );
```
xulman: assertCanCastTo().... simply to test if the returned object is really of the `PyramidalDataset` type... to make this test even stronger?

xulman: Claude suggests:

```java
 Expected e = Expected.class.cast(obj); // throws ClassCastException if wrong
```

xulman (commented):
3/N

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
xulman: `Cast.unchecked()` is actuallyu also throwing an exception.. so the same effect. CLosed.

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/PyramidContents.java
xulman: or, alternatively, we could introduce a 2nd parameter alike `ignoreFailsafeException`, and the method woud return to its previous contract (failing when no matchingf res can be found).

xulman (commented):
4/N

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/metadata/AxisCalibration.java
xulman: Please, at least adjust the name of the method so that it is apparent that it ships-in default/hardcoded/predefined/madeup.... AI suggests:

`createFixtureCalibrationForAxisNames()`
or instead of `fixture`, consider please `dummy`, `sample`, `placeholder`, `mock`

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
xulman: actually I'm wrong:
`Cast.unchecked()` will not throw the exception immediately at L377 -- at the moment of the casting. The exception will come out later, as soon as the `pyramidalDataset` is used incorrectly -- meaning, if it contains something else than `PyramidalDataset` and thus reaching out to `PyramidalDataset` API is failing.

Suggested fix is to use:

```java
PyramidalDataset pyramidalDataset = PyramidalDataset.class.cast( dataset );
```

xulman: If you @stefanhahmann do see it similarly, perhaps we shall replace all usages of `Cast.unchecked()` in our tests??

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
xulman: Aha, okay... previously we had `ImagePlus`, now it is `ImgPlus` and thus the reported dimensions is changed and so is changed this test case....

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/PyramidContents.java
xulman: > That means that callers are expected to write more lines of code anyway (to obtain an image from unknown resolution level, and to check ("do the math themselves") if there is any reasonable res

...such as `ZarrOpener#mayOpenLevel()` that is "doing the math" itself at its beginning

in fact the above proposed pattern shall be updated:

```java
int goodLevel = pc.suggestResolutionLevel( preferences );
if (goodLevel == -1) {
  if (confirmWithDialog()) return pc.asSmallestImg();
  return; //stop without opening anything
}
return pc.asImg(goodLevel);
```

xulman (commented):
N/N

I'm finally done with the review. Sorry for many, and often non-trivial suggestions.

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/ZarrUtils.java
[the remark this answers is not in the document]
stefanhahmann: Adopted the principle. `parentUri` no longer cuts up the string form, it lets URI resolution do the work:                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                  
```
   public static URI parentUri( final URI uri )                                                                                                                                                                                                                                                                   
   {                                                                                                                                                                                                                                                                                                              
       if ( uri == null )                                                                                                                                                                                                                                                                                         
           return null;                                                                                                                                                                                                                                                                                           
       final String rawPath = uri.getRawPath();                                                                                                                                                                                                                                                                   
       if ( rawPath == null ) // opaque URI: no path to walk up                                                                                                                                                                                                                                                   
           return null;                                                                                                                                                                                                                                                                                           
       if ( stripTrailingSlashes( rawPath ).lastIndexOf( '/' ) < 0 )                                                                                                                                                                                                                                              
           return null; // no path segment to remove, hence no parent                                                                                                                                                                                                                                             
       return rawPath.endsWith( "/" ) ? uri.resolve( ".." ) : uri.resolve( "." );                                                                                                                                                                                                                                 
   }  
```                                                                                                                                                                                                                                                                                                            
I went with resolve() rather than Paths.get(uri.getPath()).getParent(), because the Paths detour has potenials problems:                                                                                                                                                                         

- `getPath()` is percent-decoded, so Paths.get("/data/my data.zarr/0").getParent() gives back a string with a space, and uri.resolve(...) then throws `IllegalArgumentException: Illegal character in path at index 11`. Any dataset in a folder with a space in its name would fail to open.                      
- `getPath()` is `null` for an opaque URI, so mailto:someone@example.com would throw a NPE instead of returning null.                                                                                                                                                                                                     
- `Paths` uses the default filesystem, so on Windows the parent comes back as \data\img.ome.zarr and resolve() throws again. It also drops the trailing slash, which the callers rely on to treat the result as a folder.                                                                                        

Solution: 
`resolve(".")` avoids all three (it works on the raw path and never leaves URI-land), and resolve("..") covers the case where the dropped location already ends in a slash. The `stripTrailingSlashes` call that's left is only a read-only check for "is there a path segment to remove at all".
                                                                                                                                                                                                                                                                                                              
Behaviour, with the interesting cases (also addresses Copilot's finding. The old code turned https://example.com into https://):                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                
| input                                               | parent                        |                                                                                                                                                                                                                        
|-----------------------------------------------------|-------------------------------|                                                                                                                                                                                                                        
| file:///data/img.ome.zarr/0                         | file:/data/img.ome.zarr/      |                                                                                                                                                                                                                        
| file:///data/my%20img.ome.zarr/0                    | file:/data/my%20img.ome.zarr/ |                                                                                                                                                                                                                        
| s3://bucket/img.ome.zarr/0                          | s3://bucket/img.ome.zarr/     |                                                                                                                                                                                                                        
| https://example.com/a/b?x=1                         | https://example.com/a/        |                                                                                                                                                                                                                        
| https://example.com, https://example.com/, mailto:… | null                          |                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                               
One consequence worth flagging: resolution normalizes the empty authority, so file:///a/b → file:/a/ rather than file:///a/. Your version normalizes the same way, so I assume that's fine.

ome-zarr-fiji/src/main/java/ome/zarr/fiji/open/ZarrOpener.java
```diff
@@ -286,42 +362,25 @@ private Object openPyramidImage( final Supplier< Object > multiScaleOpener, fina
...
+	private void showSingleArrayAxesUnknown( final SingleArrayAxesUnknownException e )
 	{
-		errorHandler.accept(
-				"Opening a single resolution OME-Zarr dataset, as was found in: " + inputUri + ", is currently not supported.\n\n"
-						+ "Consider opening one level higher in the hierarchy instead." );
-		logger.info( "Opening a single resolution OME-Zarr dataset, as was found in: {}, is currently not supported.", inputUri );
+		errorHandler.accept( "Could not determine the axes of the single OME-Zarr resolution level at: " + inputUri + "\n\r\n"
+				+ "OME-Zarr v0.4 (Zarr v2) resolution levels carry no axis metadata of their own, so a single level can only "
+				+ "be interpreted via its parent group.\n\r\nConsider opening one level higher in the hierarchy instead." );
+		logger.info( "Cannot determine axes of single resolution level at {}: {}", inputUri, e.getMessage() );
```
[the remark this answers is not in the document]
stefanhahmann: Good catch. The dialog claimed a cause the exception doesn't imply.

New text:                                                                                                                                                                                                                                 
`
The array declares no axis names of its own and no parent multiscales metadata listing it could be read.
`

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
```diff
@@ -591,66 +454,54 @@ void testOpenMultiScaleDatasetBDV( String resource ) throws URISyntaxException,
...
+			assertNotNull( pyramidals );
+			// A single resolution level opens in BDV as a one-level PyramidalBdv, tracked like a multiscale image.
+			assertEquals( 1, pyramidals.size() );
+
+			PyramidalBdv< ? > pyramidalBdv = Cast.unchecked( pyramidals.get( 0 ) );
+			List< ConverterSetup > converterSetups = bdvHandle.getConverterSetups().getConverterSetups( pyramidalBdv.asSources() );
+			assertNotNull( converterSetups );
+			// One converter setup per channel; only the xyc, xyct, xyzc and 5d datasets have a c axis.
+			boolean threeChannels = resource.contains( "5d_testing" ) || resource.contains( "xyc" ) || resource.contains( "xyzc" );
+			assertEquals( threeChannels ? 3 : 1, converterSetups.size() );
```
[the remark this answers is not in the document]
stefanhahmann: False positive, since `xyc` is a prefix of / contained in `xyct`.

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/ZarrUtils.java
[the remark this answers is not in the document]
stefanhahmann: You're right about `normalize()`.
I checked: it does collapse duplicated slashes inside the path (https://h/a//b → https://h/a/b), not just ./.. segments.                                                                                                                                                       
The updated `parentUri()` now contains that: `file:///a/b//` comes out as `file:/a/`. So I didn't add a separate helper.

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/ZarrUtils.java
stefanhahmann: Partly addressed: `parentUri` no longer assembles `URI`s by cutting strings (see above). 

`lastSegment` and `isChildPath` still do string comparison and use this method. They compare against a multiscales `datasets[].path` value, which is a plain relative string from the JSON rather than a URI.

stefanhahmann: You're right about `normalize()`.
I checked: it does collapse duplicated slashes inside the path (https://h/a//b → https://h/a/b), not just ./.. segments.                                                                                                                                                       
The updated `parentUri()` now contains that: `file:///a/b//` comes out as `file:/a/`. So I didn't add a separate helper.

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/PyramidContents.java
stefanhahmann: Good catch, done as advised: `widthAtLevel()` is gone, `sizeAlongAxis( axisName, resolutionLevel )` is public, and `sizeAlongAxis( axisName )` delegates to level 0. Width is now `sizeAlongAxis( AxisCalibration.X, level )` at every call site.

One detail: the level is resolved before the absent-axis check, so an out-of-range level throws `IndexOutOfBoundsException` even when the axis is missing, otherwise `sizeAlongAxis( "z", 99 )` would silently return 1 on a 2D image.

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
stefanhahmann: Right, the backend choice was not asserted at all. Fixed: the test now reads `ZarrOpener`s private `backend` field reflectively and asserts `assertInstanceOf( N5PyramidBackend.class, ... )`, plus `assertEquals( ZarrReaderBackend.N5, ZarrOpeningSettings.DEFAULT_READER_BACKEND )`.
The two old assertions stay, but only as "the wired backend is actually usable".

Pinning `N5` explicitly means the test fails if someone flips the default.

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
stefanhahmann: Your own follow-up is right: `ImagePlus.getDimensions()` always returns five numbers (padded to xyczt), while `ImgPlus.dimensionsAsLongArray()` returns only the axes the dataset actually has. Two for a 2D image.

One more piece of context: this test was `@Disabled` before ("full support for opening single scale images is not yet implemented"), so the five-number assertion before did not run through.

ome-zarr-fiji-ui/src/test/java/ome/zarr/fijiui/open/ZarrOpenActionsTest.java
stefanhahmann: I found that

`PyramidalDataset pyramidalDataset = assertInstanceOf( PyramidalDataset.class, dataset );`

is possible instead of 

`PyramidalDataset pyramidalDataset = Cast.unchecked( dataset );`

This is doing the test for the right class and doing the actual cast at the same time. The javadoc of it:

grafik

stefanhahmann: - I changed 31 of the 42 sites to `assertInstanceOf()`
- 10 sites are in `examples/legacy` I did not want to touch them. Most likely, we can delete these folders anyway.
- 1 site was not so easily replacable because of a cast to a generic type variable

ome-zarr-imglib2/src/main/java/ome/zarr/imglib2/PyramidContents.java
stefanhahmann: - Renamed `selectResolutionLevel` → `suggestResolutionLevel`. It no longer returns the coarsest level as a "closest match" silently. When nothing fits it returns `PyramidContents.NO_MATCHING_LEVEL`
- Added `asSmallestImg()`, `asLargestImg()`, plus `smallestResolutionLevel()`
- I kept `asImg()` with a reference in the javadoc to `asLargestImg()`
- `ZarrOpener.openIJWithImage()` now follows your pattern: 
  1. suggest a level. If it fits, it is silently opened
  2. if NO_MATCHING_LEVEL: offer the coarsest one, which is opened only if the user confirms.
- I decided against the `ignoreFailsafeException` variant: the -1 / NO_MATCHING_LEVEL constant matches the existing axisIndex()   
  convention in the same class.                                                                                                               
- `mayOpenLevel` is also removed for the explicit-level path `openIJWithImage( int )`
- Tests updated in PyramidBackendTestBase (the too-narrow case now asserts NO_MATCHING_LEVEL)

PR #100, opened 2026-08-17
Run LegacyInjector.preinit() in IJ1-touching services

stefanhahmann:
Copying `ome-zarr-fiji` into a vanilla Fiji-Latest produced four `Cannot load class: ij.X` warnings at startup — `LinkageError: loader 'app' attempted duplicate class definition` for `ij.ImageJ`, `ij.ImagePlus`, `ij.gui.ImageWindow` and `ij.gui.StackWindow`. Fiji-Stable did not even start on Windows and Linux (on MacOS it still did).

Cause: `PyramidalService` is a `@Plugin` service, so SciJava creates it while the `Context` is still being built and reflects over its declared fields and methods. That resolves the `ij.*` types in its signatures and loaded them before `LegacyService.<clinit>` had run `LegacyInjector.preinit()`, so the patcher could no longer redefine them and they ended up without their IJ1/IJ2 bridge hooks.

Calling `preinit()` from a static initializer patches IJ1 first.

`PasteOmeZarrUriToolInstaller` worked around the same problem by routing all `ij.*` access through `Class.forName`; that is no longer needed, so it now uses plain typed `Toolbar` calls.

Its `@EventHandler(UIShownEvent)` deferral stays (`Toolbar.getInstance()` is `null` until the IJ1 UI is up), as does the reflective write to the private `Toolbar.installingStartupTool`.

No new jar to install: every Fiji ships `ij1-patcher`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/plugin/PasteOmeZarrUriToolInstaller.java
+++ b/ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/plugin/PasteOmeZarrUriToolInstaller.java
@@ -34,12 +34,12 @@ import java.awt.KeyboardFocusManager;
 import java.awt.event.KeyEvent;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Field;
-import java.lang.reflect.Method;
 import java.net.URI;
 
 import javax.swing.SwingUtilities;
 import javax.swing.text.JTextComponent;
 
+import net.imagej.patcher.LegacyInjector;
 import org.scijava.Context;
 import org.scijava.event.EventHandler;
 import org.scijava.plugin.Plugin;
@@ -51,6 +51,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import ij.IJ;
+import ij.gui.Toolbar;
 import ij.plugin.tool.MacroToolRunner;
 import ome.zarr.fijiui.open.PasteToOpenAction;
 import ome.zarr.fijiui.util.ClipboardUtils;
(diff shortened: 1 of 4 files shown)


ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/plugin/PasteOmeZarrUriToolInstaller.java
```diff
@@ -105,6 +102,13 @@
 @Plugin( type = Service.class )
```
xulman, line 102: Please, use `SciJavaService` instead of `Service`.

It changes nothing as the former is no-adding derived from the latter, but it
- looks better :-) and less general,
- and is consistent with `PyramidalService.java`

ome-zarr-fiji/src/main/java/ome/zarr/fiji/plugins/PyramidalService.java
```diff
@@ -62,6 +63,13 @@
 @Plugin( type = SciJavaService.class )
 public class PyramidalService extends AbstractService implements SciJavaService
 {
+	static
```
xulman, line 66: After some independent investigation of mine of this regard, this indeed seems to be the best solution (as things in Fiji land stand today).

However, I'm not sure about the other instance/addition of exactly this same code snippet. See the other comment.

ome-zarr-fiji/pom.xml
```diff
@@ -117,6 +117,10 @@
 			<groupId>net.imagej</groupId>
 			<artifactId>imagej-common</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>net.imagej</groupId>
+			<artifactId>ij1-patcher</artifactId>
+		</dependency>
```
xulman, line 123: Okay, really, it has to be here, good.

(I thought it comes transitively with some other `net.image` artifacts, and apparently I was wrong.)

ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/plugin/PasteOmeZarrUriToolInstaller.java
```diff
@@ -105,6 +102,13 @@
...
 public class PasteOmeZarrUriToolInstaller extends AbstractService implements SciJavaService
 {
+	static
```
xulman, line 105: I guess this is only here to deal with "racing conditions" on `SciJavaServices` initialization?
That said, if by chance this service object would be constructed earlier than that of `PyramidalService`, because Fiji itself probably doens't care about dependecies of artifacts and doesn't initialize "bottom-to-top" (this package is directly depending on another package that includes `PyramidalService`).

...cause in my local testing, this code needs not be here....

xulman: Aaah, it is because of the now-explicit `Toolbar` class...

xulman: Tested w/o the `LegacyInjector.preinit();`, and it is happy without it (Fiji starts nicely).

It feels redundant to have it here, but it is
- one-time/start-up cost -> no big deal then
- "innocent fixer" (that's likely already at several place in Fiji ecosystem).

...so I'm okay to live with it here.

Also the code clean up below is fine, and welcome.

xulman (approved):
Honestly, I wanted to do everything else than patching Fiji ourselves with the `LegacyInjector.preinit();` but it is nowadays pretty much a standard proceedure :-/
so we can have it too...

Kudos to AI that it has noticed the cosmetic cleanup in the `PasteOmeZarrUriTooInstaller`.

Please fix (also cosmetics!) the `Service` to `SciJavaService` to have the code base even nicer :)

ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/plugin/PasteOmeZarrUriToolInstaller.java
stefanhahmann: Done.

ome-zarr-fiji/src/main/java/ome/zarr/fiji/plugins/PyramidalService.java
stefanhahmann: Ok.

ome-zarr-fiji/pom.xml
stefanhahmann: I checked this.

So far `ij1-patcher` only came as test scope dependency, but we need it now at compile time as well. In the `ome-zarr-fiji` artefact, it so far did not come as compile time dependency:

grafik

ome-zarr-fiji-ui/src/main/java/ome/zarr/fijiui/plugin/PasteOmeZarrUriToolInstaller.java
stefanhahmann: Yes, I think the reason to have it here again is to guard against the case that the `PasteOmeZarrUriToolInstaller` `SciJavaService` is loaded before the `PyramidalService`, which could (by chance) happen.

We are not alone in the FIJI world using this pattern: 
https://github.com/search?q=LegacyInjector.preinit%28%29+language%3AJava&type=code&l=Java

PR #101, opened 2026-08-18
Add LinkHandler plugin for fiji:// URIs

stefanhahmann:
Adds `LinkHandlerPlugin`, a SciJava `LinkHandler` that opens OME-Zarr datasets from `fiji://open?...zarr...` URIs.

Original work by @xulman on [`fiji_links`](https://github.com/BioImageTools/ome-zarr-fiji-java/tree/fiji_links) (`0676a38c`), rebased here onto `main` to pick up the many changes in the meantime.

**Note:** `scijava-desktop` is built for Java 11, so `ome-zarr-fiji-ui` now overrides `scijava.jvm.version` from `ome-zarr-parent` `pom.xml` to 11 and no longer excludes it. The other modules stay on Java 8. This would mean the `ome-zarr-fiji-ui` artifact would not run on a Java 8 Fiji-Stable.

There are yet no tests for the handler.


stefanhahmann:
Superseded by #102, which takes a much smaller route.

Fiji-Latest already ships `sc.fiji:fiji-links`, whose `OpenLinkHandler` owns the `fiji://open/{file,url,source}?p=…` syntax, the OS-level scheme registration and the URI parsing, and finishes by calling `IOService.open(Location)`. That dispatches to whichever `IOPlugin` claims the location — i.e. to our existing DnD plugin. So `fiji://open/file?p=…` already worked without this PR; only the remote case was missing, and closing that gap took removing one `instanceof FileLocation` guard.

Suggest closing this one in favour of #102, unless there is something here it misses.

stefanhahmann:
Close in favor of #102

